PR #77, opened 2022-08-09
[Governance] unify table UI for Transactions and Proposals

zihan-aptos:
## Nothing changed for Transactions table:
Screen Shot 2022-08-08 at 5 56 27 PM
## Apply same UI design for Proposals table:
Screen Shot 2022-08-08 at 5 56 43 PM

### DIFF
--- a/src/components/HashButton.tsx
+++ b/src/components/HashButton.tsx
@@ -1,0 +1,129 @@
+import React from "react";
+import {
+  Box,
+  Button,
+  Typography,
+  Popover,
+  IconButton,
+  useTheme,
+} from "@mui/material";
+import {grey} from "../themes/colors/aptosColorPalette";
+import ChevronRightRoundedIcon from "@mui/icons-material/ChevronRightRounded";
+import ChevronLeftRoundedIcon from "@mui/icons-material/ChevronLeftRounded";
+import {truncateAddress} from "../pages/utils";
+
+interface HashButtonProps {
+  hash: string;
+}
+
+export default function HashButton({hash}: HashButtonProps) {
+  const [anchorEl, setAnchorEl] = React.useState<HTMLButtonElement | null>(
+    null,
+  );
+
+  const hashExpand = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.stopPropagation();
+    setAnchorEl(event.currentTarget);
+  };
+
+  const hashCollapse = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.stopPropagation();
+    setAnchorEl(null);
+  };
+
+  const open = Boolean(anchorEl);
+  const id = open ? "simple-popover" : undefined;
+  const theme = useTheme();
+
+  return (
+    <Box>
+      <Button
+        sx={{
+          textTransform: "none",
+          backgroundColor: `${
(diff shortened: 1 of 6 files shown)


src/components/WalletButton.tsx
```diff
@@ -32,11 +32,19 @@ const WalletButtonWrapper = ({
   return (
     <Button
       variant="outlined"
-      sx={{padding: "10px 25px", backgroundColor: theme.palette.mode === "dark" ? "#1B1F1E" : "white",color: theme.palette.mode === "dark" ? "#2ED8A7" : "#121615"}}
+      sx={{
+        padding: "10px 25px",
+        backgroundColor: theme.palette.mode === "dark" ? "#1B1F1E" : "white",
```
zacharydenton, line 37: Can it use colors from the global color palette?

src/components/WalletButton.tsx
```diff
@@ -32,11 +32,19 @@ const WalletButtonWrapper = ({
...
+        color: theme.palette.mode === "dark" ? "#2ED8A7" : "#121615",
+      }}
       {...props}
     >
       {icon}
-      <Typography variant="body2" color={theme.palette.mode === "dark" ? "white" : "121615"} ml={2}>
+      <Typography
+        variant="body2"
+        color={theme.palette.mode === "dark" ? "white" : "121615"}
```
zacharydenton, line 45: `"121615"` is not a valid color.

zacharydenton (approved):

src/components/HashButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+  hash: string;
+}
+
+export default function HashButton({hash}: HashButtonProps) {
+  const [anchorEl, setAnchorEl] = React.useState<HTMLButtonElement | null>(
+    null,
+  );
+
+  const hashExpand = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.stopPropagation();
```
zacharydenton, line 25: Can we redo it with preventDefault() / defaultPrevented instead of stopPropagation()? https://css-tricks.com/dangers-stopping-event-propagation/

src/components/WalletButton.tsx
zihan-aptos: I'll leave this change to @msavir for #78 to avoid merge conflicts.

src/components/WalletButton.tsx
zihan-aptos: I'll leave this change to @msavir for #78 to avoid merge conflicts.

msavir: @zihan-aptos could you remove this file from the PR then?

msavir (approved):
lgtm

src/components/HashButton.tsx
zihan-aptos: I tried `preventDefault()`. However, it does not work. Seems like `preventDefault()` does not stop the event from bubbling up, so `rowClick` will be called in the row component. Maybe @moonclavedev has a better idea since he authored this part of code.

zacharydenton: Yeah it would require a check for event.defaultPrevented in the event handler for the row navigation. No big deal

PR #78, opened 2022-08-09
create stake BCS transaction

msavir:
This PR 

1. installs `aptos` version 1.2.2 to match with the wallet version
2. removes the stakeEndTime from stake page UI as it is not needed anymore
3. implements the stake transaction
4. switch JSON transactions into BCS transactions
5. creates a global submitTransaction hook that accepts a transaction payload and handles the transaction response state

### DIFF
--- a/src/api/hooks/useSubmitTransaction.ts
+++ b/src/api/hooks/useSubmitTransaction.ts
@@ -2,7 +2,9 @@ import {TxnBuilderTypes} from "aptos";
 import {useState} from "react";
 import {signAndSubmitTransaction} from "../wallet";
 
-export type TransactionResponse = TransactionResponseOnSuccess | TransactionResponseOnFailure;
+export type TransactionResponse =
+  | TransactionResponseOnSuccess
+  | TransactionResponseOnFailure;
 
 export type TransactionResponseOnSuccess = {
   succeeded: boolean;
@@ -15,12 +17,11 @@ export type TransactionResponseOnFailure = {
 };
 
 const useSubmitTransaction = () => {
-  const [transactionResponse, setTransactionResponse] = useState<TransactionResponse | null>(
-    null,
-  );
+  const [transactionResponse, setTransactionResponse] =
+    useState<TransactionResponse | null>(null);
 
   async function submitTransaction(
-    payload: TxnBuilderTypes.TransactionPayloadScriptFunction
+    payload: TxnBuilderTypes.TransactionPayloadScriptFunction,
   ) {
     await signAndSubmitTransaction(payload).then(setTransactionResponse);
   }
--- a/src/api/hooks/useSubmitTransaction.ts
+++ b/src/api/hooks/useSubmitTransaction.ts
@@ -1,0 +1,35 @@
+import {TxnBuilderTypes} from "aptos";
+import {useState} from "react";
+import {signAndSubmitTransaction} from "../wallet";
+
+export type TransactionResponse = TransactionResponseOnSuccess | TransactionResponseOnFailure;
+
+export type TransactionResponseOnSuccess = {
+  succeeded: boolean;
+  transactionHash: string;
+};
+
+export type TransactionResponseOnFailure = {
+  succeeded: boolean;
+  message: string;
+};
+
+const useSubmitTransaction = () => {
+  const [transactionResponse, setTransactionResponse] = useState<TransactionResponse | null>(
+    null,
+  );
+
+  async function submitTransaction(
+    payload: TxnBuilderTypes.TransactionPayloadScriptFunction
+  ) {
+    await signAndSubmitTransaction(payload).then(setTransactionResponse);
+  }
+
+  function clearTransactionResponse() {
+    setTransactionResponse(null);
+  }
+
+  return {submitTransaction, transactionResponse, clearTransactionResponse};
+};
+
+export default useSubmitTransaction;
--- a/src/api/hooks/useSubmitTransaction.ts
+++ b/src/api/hooks/useSubmitTransaction.ts
@@ -7,12 +7,12 @@ export type TransactionResponse =
   | TransactionResponseOnFailure;
 
 export type TransactionResponseOnSuccess = {
-  succeeded: boolean;
+  succeeded: true;
   transactionHash: string;
 };
 
 export type TransactionResponseOnFailure = {
-  succeeded: boolean;
+  succeeded: false;
   message: string;
 };
 
--- a/src/components/WalletButton.tsx
+++ b/src/components/WalletButton.tsx
@@ -32,11 +32,19 @@ const WalletButtonWrapper = ({
   return (
     <Button
       variant="outlined"
-      sx={{padding: "10px 25px", backgroundColor: theme.palette.mode === "dark" ? "#1B1F1E" : "white",color: theme.palette.mode === "dark" ? "#2ED8A7" : "#121615"}}
+      sx={{
+        padding: "10px 25px",
+        backgroundColor: theme.palette.mode === "dark" ? "#1B1F1E" : "white",
+        color: theme.palette.mode === "dark" ? "#2ED8A7" : "#121615",
+      }}
       {...props}
     >
       {icon}
-      <Typography variant="body2" color={theme.palette.mode === "dark" ? "white" : "121615"} ml={2}>
+      <Typography
+        variant="body2"
+        color={theme.palette.mode === "dark" ? "white" : "121615"}
+        ml={2}
+      >
         {text}
       </Typography>
       {children}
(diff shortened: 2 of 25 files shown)


src/api/hooks/useSubmitTransaction.ts
```diff
@@ -0,0 +1,36 @@
+import {TxnBuilderTypes} from "aptos";
+import {useState} from "react";
+import {signAndSubmitTransaction} from "../wallet";
+
+export type TransactionResponse =
+  | TransactionResponseOnSuccess
+  | TransactionResponseOnFailure;
+
+export type TransactionResponseOnSuccess = {
+  succeeded: boolean;
```
zacharydenton, line 10: Is it possible for succeeded to be false here? I think the type could be `succeeded: true`?

src/components/WalletButton.tsx
```diff
@@ -32,11 +32,19 @@ const WalletButtonWrapper = ({
   return (
     <Button
       variant="outlined"
-      sx={{padding: "10px 25px", backgroundColor: theme.palette.mode === "dark" ? "#1B1F1E" : "white",color: theme.palette.mode === "dark" ? "#2ED8A7" : "#121615"}}
+      sx={{
+        padding: "10px 25px",
+        backgroundColor: theme.palette.mode === "dark" ? "#1B1F1E" : "white",
```
zacharydenton, line 37: Is it possible to use color values from the global color palette?

src/components/WalletButton.tsx
```diff
@@ -32,11 +32,19 @@ const WalletButtonWrapper = ({
...
+        color: theme.palette.mode === "dark" ? "#2ED8A7" : "#121615",
+      }}
       {...props}
     >
       {icon}
-      <Typography variant="body2" color={theme.palette.mode === "dark" ? "white" : "121615"} ml={2}>
+      <Typography
+        variant="body2"
+        color={theme.palette.mode === "dark" ? "white" : "121615"}
```
zacharydenton, line 45: `"121615"` is not a valid color

src/pages/Governance/Proposal/card/VoteButtons.tsx
```diff
@@ -110,7 +111,7 @@ export default function VoteButtons({proposalId}: Props) {
         action={voteOnSuccessSnackbarAction}
       >
         {`Vote succeeded with transaction ${
-          (voteResponse as VoteResponseOnSuccess).transactionHash
+          (transactionResponse as TransactionResponseOnSuccess).transactionHash
```
zacharydenton, line 114: It should be possible to omit the `as TransactionResponseOnSuccess` cast (possibly in combination with the change to `succeeded: true / false` mentioned above). https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing It's best to avoid using `as` because it can cause things to break at runtime (https://dev.to/ylerjen/typescript-cast-is-a-type-breaker-1jbh has more details).

src/pages/Governance/Stake/Index.tsx
```diff
@@ -38,68 +43,133 @@ export function StakePage() {
   };
 
   const onSubmitClick = async () => {
-    // TODO: to be implemented
-    console.log("Submit Staking");
+    if (!stakingAmount) return;
+    await submitStake(parseInt(stakingAmount), operatorAddr, voterAddr);
```
zacharydenton, line 47: How come the stakingAmount is stored as a string and not `number | null`? Keep in mind that parseInt can return `NaN` if the string isn't valid. I would suggest handling the input validation and conversion to integer in `onStakingAmountChange`.

src/pages/Governance/hooks/useSubmitStake.ts
```diff
@@ -0,0 +1,32 @@
...
+const useSubmitStake = () => {
+  const {submitTransaction, transactionResponse, clearTransactionResponse} =
+    useSubmitTransaction();
+
+  async function submitStake(
+    stakingAmount: number,
+    operatorAddr: string,
+    voterAddr: string,
+  ) {
+    const payload = new TxnBuilderTypes.TransactionPayloadScriptFunction(
```
zacharydenton, line 13: I think the function could use some validation before submitting the transaction (e.g. stakingAmount is an integer in a reasonable range, operatorAddr and voterAddr are valid addresses, etc.).

src/pages/Governance/hooks/useSubmitVote.ts
```diff
@@ -1,67 +1,36 @@
...
+const useSubmitVote = () => {
+  const {submitTransaction, transactionResponse, clearTransactionResponse} =
+    useSubmitTransaction();
+
+  async function submitVote(
+    proposalId: number,
+    shouldPass: boolean,
+    ownerAccountAddr: string,
+  ) {
+    let serializer = new BCS.Serializer();
```
zacharydenton, line 13: Any reason not to use `const`?

zacharydenton (approved):

src/pages/Governance/Stake/Index.tsx
```diff
@@ -38,68 +43,133 @@ export function StakePage() {
...
+
+  useEffect(() => {
+    if (transactionResponse?.succeeded) {
+      setStakingAmount("");
+      setOperatorAddr("");
+      setVoterAddr("");
+    }
+  }, [transactionResponse]);
+
+  const voteOnSuccessSnackbarAction = (
```
zihan-aptos, line 62: `stakeOnSuccess...`?

src/pages/Governance/Stake/Index.tsx
```diff
@@ -38,68 +43,133 @@ export function StakePage() {
...
+      size="small"
+      aria-label="close"
+      color="inherit"
+      onClick={onCloseErrorAlert}
+    >
+      <CloseIcon fontSize="small" />
+    </IconButton>
+  );
+
+  const voteOnSuccessSnackbar = transactionResponse !== null && (
```
zihan-aptos, line 73: naming

src/pages/Governance/Stake/Index.tsx
```diff
@@ -38,68 +43,133 @@ export function StakePage() {
...
+        action={voteOnSuccessSnackbarAction}
+      >
+        {`Stake succeeded with transaction ${
+          (transactionResponse as TransactionResponseOnSuccess).transactionHash
+        }.`}
+      </Alert>
+    </Snackbar>
+  );
+
+  const voteOnFailureSnackbarAction = (
```
zihan-aptos, line 93: ^

src/pages/Governance/Stake/Index.tsx
```diff
@@ -38,68 +43,133 @@ export function StakePage() {
...
+      size="small"
+      aria-label="close"
+      color="inherit"
+      onClick={onCloseErrorAlert}
+    >
+      <CloseIcon fontSize="small" />
+    </IconButton>
+  );
+
+  const voteOnFailureSnackbar = transactionResponse !== null && (
```
zihan-aptos, line 104: ^

src/pages/Governance/Stake/Index.tsx
```diff
@@ -38,68 +43,133 @@ export function StakePage() {
...
-            <LocalizationProvider dateAdapter={AdapterDateFns}>
-              <DateTimePicker
-                label="Staking End Time"
-                value={stakingEndTime}
-                onChange={onStakingEndTimeChange}
-                renderInput={(params) => <TextField {...params} />}
-                minTime={new Date()}
+    <>
+      {voteOnSuccessSnackbar}
+      {voteOnFailureSnackbar}
```
zihan-aptos, line 127: Seems like we're reusing the snack bar from the vote buttons. Not a blocker, but let's modularize this part and make it reusable. I can help with this, let me know.

zihan-aptos (approved):

src/pages/Governance/Proposal/card/VoteButtons.tsx
msavir: I tried to change to `succeeded: true / false` - that doesnt seem to help. We need to specify what type to use since this is a union type and not all types have the same props. (looks like the article you attached is also ok with precising the type from union types)

src/pages/Governance/Stake/Index.tsx
msavir: we can't set a HTML input to null (getting console error) and we can't set it to undefined as then we get a changing an uncontrolled input to be controlled error.
I added a validation to replace any non-digit to un empty string and then parse the value - can go back to it later if you feel we should do it differently.

src/pages/Governance/Stake/Index.tsx
msavir: agree. was thinking on doing it in a separate PR to keep this one light. feel free to tackle it =)

src/pages/Governance/hooks/useSubmitStake.ts
msavir: added on the `onSubmitClick()`

PR #80, opened 2022-08-09
[Governance] proposal status

zihan-aptos:
* proposals table: add icon for status column
* proposal page header
    * add icon for status
    * hide "Time Remaining" when the proposal is closed
* proposal page 
    * hide vote buttons when the proposal is closed
Screen Shot 2022-08-09 at 12 17 43 PM
Screen Shot 2022-08-09 at 12 18 33 PM
Screen Shot 2022-08-09 at 12 18 44 PM
Screen Shot 2022-08-09 at 12 19 25 PM

### DIFF
--- a/src/pages/Governance/Proposal/Header.tsx
+++ b/src/pages/Governance/Proposal/Header.tsx
@@ -8,7 +8,7 @@ import PersonIcon from "@mui/icons-material/Person";
 import AccessTimeIcon from "@mui/icons-material/AccessTime";
 import {useTheme} from "@mui/material/styles";
 import useMediaQuery from "@mui/material/useMediaQuery";
-import {renderStatusIcon, isVotingClosed} from "../utils";
+import {getStatusColor, renderStatusIcon, isVotingClosed} from "../utils";
 
 const SECONDARY_TEXT_COLOR = "#A3A3A3";
 const HASH_WIDTH = 200;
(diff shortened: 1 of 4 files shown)


src/pages/Governance/utils.tsx
```diff
@@ -44,3 +48,41 @@ export function getProposalState(proposal: Proposal): ProposalState {
...
   }
 }
+
+export function renderStatusIcon(proposalState: ProposalState) {
+  switch (proposalState) {
+    case ProposalState.SUCCEEDED:
+      return (
+        <CheckCircleOutlinedIcon
+          fontSize="small"
+          // color={primaryColor}
```
msavir, line 58: can we remove those comments?

src/pages/Governance/Proposal/Header.tsx
```diff
@@ -47,6 +50,10 @@ function TimeRemainingComponent({
   proposal: Proposal;
   isOnMobile: boolean;
 }) {
+  if (isVotingClosed(proposal)) {
```
msavir, line 53: not sure if we dont want to show anything... maybe we can display the time it was closed?

zihan-aptos: I thought about it but with the current design there isn't a good way to display the time it was closed so I just hide it for simplicity. Let me add a TODO here and revisit after consulting Josh. Btw do you know what is the accurate "close time"?

src/pages/Governance/Proposal/card/Index.tsx
```diff
@@ -22,7 +23,9 @@ export function ProposalCard({proposal}: ProposalCardProps) {
           />
         }
       >
-        <CastVoteSection proposalId={proposal.proposal_id} />
+        {!isVotingClosed(proposal) && (
```
msavir, line 26: do we also want to hide the vote option when the user has already voted?

msavir (approved):
lgtm

PR #81, opened 2022-08-09
[GOVERNANCE] handle create account event

msavir:
This PR sets the account created state and lets the user know if they need to create an account by displaying it on the wallet button or the connect step in the instructions section on the proposals page

https://user-images.githubusercontent.com/29798064/183765092-b7d2f965-26a8-4792-bee9-63c4d06c153c.mov

### DIFF
--- a/src/components/WalletButton.tsx
+++ b/src/components/WalletButton.tsx
@@ -12,6 +12,7 @@ import {
 } from "@mui/material";
 import {useWalletContext} from "../context/wallet/context";
 import CreditCardIcon from "@mui/icons-material/CreditCard";
+import AddCardIcon from "@mui/icons-material/AddCard";
 import ErrorIcon from "@mui/icons-material/Error";
 import {truncateAddress} from "../pages/utils";
 import {isUpdatedVersion} from "../api/wallet";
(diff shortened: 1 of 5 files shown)


src/components/WalletButton.tsx
```diff
@@ -99,7 +100,20 @@ export const WalletButton = (): JSX.Element => {
 
   return (
     <>
-      {isInstalled && !isConnected && (
+      {isInstalled && !isAccountSet && (
+        <Tooltip title="Use the Wallet extension to create an account">
+          <span>
+            <WalletButtonWrapper
+              text="create an account"
```
zihan-aptos, line 107: Do we always want users to create new accounts? What if they simply want to login existing accounts?

msavir: they will open the wallet extension and have the option to do it there, they are still creating an account but by importing a private key (and not creating a **new** account)

zihan-aptos: Not a blocker, but I'm wondering if "Create an account" is confusing for some users if they want to login the wallet with private key or mnemonic.

zihan-aptos (approved):

PR #82, opened 2022-08-10
fix for when window.aptos is undefined

msavir:
fix for sentry error `Cannot read properties of null` due to window.aptos is not ready yet

https://sentry.io/organizations/aptos-labs/issues/3488980419/?project=6249755&referrer=slack

### DIFF
--- a/src/api/wallet.ts
+++ b/src/api/wallet.ts
@@ -10,7 +10,7 @@ export const getAptosWallet = (): boolean => {
 
 export const isWalletConnected = async (): Promise<boolean> => {
   try {
-    if (window.aptos && (await window.aptos.isConnected())) {
+    if (await window.aptos?.isConnected()) {
       return true;
     }
   } catch (error) {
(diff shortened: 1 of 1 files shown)


zihan-aptos (approved):

src/api/wallet.ts
```diff
@@ -44,7 +42,7 @@ export const connectToWallet = async (): Promise<boolean> => {
 
 export const getAccountAddress: () => Promise<string | null> = async () => {
   try {
-    const data = await window.aptos.account();
+    const data = await window.aptos?.account();
```
zacharydenton, line 45: The linked Sentry issue includes an error log `window.aptos.account is not a function` implying that `window.aptos` is defined but that `window.aptos.account` is not a function. I'm not sure why that would be the case but the current implementation will not handle it; it needs to be `window.aptos?.account?.()`. Potentially all of the window.aptos method calls need to be changed in this way.

msavir: it looks like the exception in this error is `TypeError: Cannot read properties of null (reading 'disconnect')` - which is a bit weird because we dont have `disconnect` in our app. 
Not completely sure what causes `window.aptos.account is not a function` but I guess to be safe we can add this check also

PR #83, opened 2022-08-10
[Governance] inputs validation

zihan-aptos:
https://user-images.githubusercontent.com/109111707/184001296-51dec86a-284b-4c57-a44a-9e613f5ca60f.mov

### DIFF
--- a/src/pages/Governance/Proposal/card/VoteButtons.tsx
+++ b/src/pages/Governance/Proposal/card/VoteButtons.tsx
@@ -1,12 +1,4 @@
-import {
-  Button,
-  Stack,
-  TextField,
-  Snackbar,
-  Alert,
-  IconButton,
-  Box,
-} from "@mui/material";
+import {Button, Stack, Snackbar, Alert, IconButton, Box} from "@mui/material";
 import React, {useState, useEffect} from "react";
 import useSubmitVote from "../../hooks/useSubmitVote";
 import CheckCircleOutlinedIcon from "@mui/icons-material/CheckCircleOutlined";
@@ -26,29 +18,29 @@ import {
   TransactionResponseOnFailure,
   TransactionResponseOnSuccess,
 } from "../../../../api/hooks/useSubmitTransaction";
-import {isValidAccountAddress} from "../../../utils";
+import useAddressInput from "../../../../api/hooks/useAddressInput";
 
 // TODO:
 // 1. check if voted
 // 2. check if eligible to vote
 
-type Props = {
+type VoteButtonsProps = {
   proposalId: string;
 };
 
-export default function VoteButtons({proposalId}: Props) {
-  const [accountAddr, setAccountAddr] = useState<string>("");
-  const [accountAddrIsValid, setAccountAddrIsValid] = useState<boolean>(true);
+export default function VoteButtons({proposalId}: VoteButtonsProps) {
   const [voteForModalIsOpen, setVoteForModalIsOpen] = useState<boolean>(false);
   const [voteAgainstModalIsOpen, setVoteAgainstModalIsOpen] =
     useState<boolean>(false);
 
+  const {
+    addr: accountAddr,
+    renderAddressTextField,
+    validateAddressInput,
+  } = useAddressInput();
+
   const {submitVote, transactionResponse, clearTransactionResponse} =
     useSubmitVote();
-
-  useEffect(() => {
-    setAccountAddrIsValid(true);
-  }, [accountAddr]);
 
   useEffect(() => {
     if (transactionResponse !== null) {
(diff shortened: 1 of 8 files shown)


src/pages/Governance/Stake/Index.tsx
```diff
@@ -44,20 +53,20 @@ export function StakePage() {
...
 
-    await submitStake(stakingAmountNumber, operatorAddr, voterAddr);
+    const isOperatorAddrValid = isValidAccountAddress(operatorAddr);
+    setOperatorAddrIsValid(isOperatorAddrValid);
+
+    const isVoterAddrValid = isValidAccountAddress(voterAddr);
+    setVoterAddrIsValid(isVoterAddrValid);
+
+    if (isOperatorAddrValid && isVoterAddrValid) {
+      await submitStake(parseInt(stakingAmount), operatorAddr, voterAddr);
```
msavir, line 68: why did you switch it and not using `stakingAmountNumber`?

src/pages/Governance/Stake/Index.tsx
```diff
@@ -24,11 +23,21 @@ import {isValidAccountAddress} from "../../utils";
...
 
   const {submitStake, transactionResponse, clearTransactionResponse} =
     useSubmitStake();
 
+  useEffect(() => {
+    setOperatorAddrIsValid(true);
+  }, [operatorAddr]);
+
+  useEffect(() => {
+    setVoterAddrIsValid(true);
```
msavir, line 38: not completely sure why you set it to `true` every time the `voterAddr` is changed? same goes to the upper useEffect and the one on `VoteButtons` file

src/pages/Governance/Proposal/card/VoteButtons.tsx
```diff
@@ -151,15 +162,29 @@ export default function VoteButtons({proposalId}: Props) {
...
-        <TextField
-          fullWidth
-          label="Owner Account Address"
-          variant="filled"
-          size="small"
-          margin="normal"
-          value={accountAddr}
-          onChange={onAccountAddrChange}
-        />
+        {accountAddrIsValid ? (
```
msavir, line 165: can we globalize those validation TextField ? looks like we use it here and also on the Stake page

src/pages/Governance/Stake/Index.tsx
```diff
@@ -44,20 +53,20 @@ export function StakePage() {
...
-    if (
-      !isValidAccountAddress(operatorAddr) ||
-      !isValidAccountAddress(voterAddr)
-    ) {
-      return;
-    }
     // TODO - this will likely be very dependent per network
     // pull from on chain config 0x1::stake::ValidatorSetConfiguration
     const stakingAmountNumber = parseInt(stakingAmount);
     if (stakingAmountNumber <= 0) return;
```
msavir, line 59: can we also show the error for the staking amount?

msavir (changes requested):

src/pages/utils.tsx
```diff
@@ -131,5 +131,5 @@ export function isValidAccountAddress(accountAddr: string): boolean {
   // account address is 0x{64 hex characters}
   // with multiple options - 0X, 0x001, 0x1, 0x01
   // can start with that and see if any fails to parsing
-  return /^(0[xX])?[a-fA-F0-9]{1,64}$/.test(accountAddr);
+  return /^(0[xX])?[a-fA-F0-9]{64}$/.test(accountAddr);
```
zacharydenton, line 134: I believe leading zeros are truncated from addresses, so the previous regex was correct. https://aptos.dev/guides/guide-for-system-integrators/#account-identifiers

src/pages/Governance/Proposal/card/VoteButtons.tsx
zihan-aptos: I didn't globalize them because they had different variant/size... But on a second thought, I might be good to make them the same variant/size to be consistent. I'll make the change.

src/pages/Governance/Stake/Index.tsx
zihan-aptos: because once user starts to change the input (typing or copy-pasting new things), we want to make the error message disappear. The idea is to only validate input when the user tries to submit a vote.

src/pages/Governance/Stake/Index.tsx
zihan-aptos: With the current check `setStakingAmount(event.target.value.replace(/[^0-9]/g, ""));`, I don't think there will ever be negative number inputs. But I'll add error message for `0`. Good catch!

msavir: also when the staking amount is left empty, `!stakingAmount` should handle both cases I believe

msavir (approved):

PR #84, opened 2022-08-10
[GOVERNANCE] check wallet and explorer use the same network

msavir:
Before sign and submit a transaction, we want to make sure the explorer and the wallet are set to the same network. 

This PR blocks and returns an error when users try to sign and submit a transaction if the explorer an wallet are not set to the same network

### DIFF
--- a/src/GlobalState.tsx
+++ b/src/GlobalState.tsx
@@ -1,21 +1,21 @@
 import React from "react";
-import {networks, defaultNetworkName, ExplorerNetworks} from "./constants";
+import {networks, defaultNetworkName, NetworkName} from "./constants";
 
-const selected_network = safeGetSelectedNetworkName() as ExplorerNetworks;
+const selected_network = safeGetSelectedNetworkName() as NetworkName;
 
-function safeGetSelectedNetworkName(): string {
+function safeGetSelectedNetworkName(): NetworkName {
   let selected_network = localStorage.getItem("selected_network");
   if (selected_network) {
     selected_network = selected_network.toLowerCase();
     if (selected_network in networks) {
-      return selected_network;
+      return selected_network as NetworkName;
     }
   }
-  return defaultNetworkName.toLowerCase();
+  return defaultNetworkName.toLowerCase() as NetworkName;
 }
 
 export type GlobalState = {
-  network_name: ExplorerNetworks;
+  network_name: NetworkName;
   network_value: string;
 };
 
--- a/src/GlobalState.tsx
+++ b/src/GlobalState.tsx
@@ -1,7 +1,7 @@
 import React from "react";
-import {networks, defaultNetworkName} from "./constants";
+import {networks, defaultNetworkName, ExplorerNetworks} from "./constants";
 
-const selected_network = safeGetSelectedNetworkName();
+const selected_network = safeGetSelectedNetworkName() as ExplorerNetworks;
 
 function safeGetSelectedNetworkName(): string {
   let selected_network = localStorage.getItem("selected_network");
@@ -15,7 +15,7 @@ function safeGetSelectedNetworkName(): string {
 }
 
 export type GlobalState = {
-  network_name: string;
+  network_name: ExplorerNetworks;
   network_value: string;
 };
 
--- a/src/GlobalState.tsx
+++ b/src/GlobalState.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import {networks, defaultNetworkName, NetworkName} from "./constants";
 
-const selected_network = safeGetSelectedNetworkName() as NetworkName;
+const selected_network = safeGetSelectedNetworkName();
 
 function safeGetSelectedNetworkName(): NetworkName {
   let selected_network = localStorage.getItem("selected_network");
@@ -11,7 +11,7 @@ function safeGetSelectedNetworkName(): NetworkName {
       return selected_network as NetworkName;
     }
   }
-  return defaultNetworkName.toLowerCase() as NetworkName;
+  return defaultNetworkName;
 }
 
 export type GlobalState = {
--- a/src/api/hooks/useSubmitTransaction.ts
+++ b/src/api/hooks/useSubmitTransaction.ts
@@ -1,7 +1,10 @@
 import {TxnBuilderTypes} from "aptos";
 import {useState} from "react";
-import {walleExplorertNetworkMap} from "../../constants";
-import {useWalletContext} from "../../context/wallet/context";
+
+import {
+  useWalletContext,
+  walletExplorertNetworkMap,
+} from "../../context/wallet/context";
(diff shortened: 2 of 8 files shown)


src/api/wallet.ts
```diff
@@ -50,6 +50,15 @@ export const getAccountAddress: () => Promise<string | null> = async () => {
   return null;
 };
 
+export const getWalletNetwork: () => Promise<string> = async () => {
```
zacharydenton, line 53: Instead of returning a string, it should return a member of an enum of valid wallet networks (i.e. `'Devnet' | 'Testnet' | 'Localhost'`).

src/constants.tsx
```diff
@@ -8,6 +8,12 @@ export const networks: Record<string, string> = {
   //governance: "https://ait2.aptosdev.com/", // TODO - use the proposals url
 };
 
+export const walletNetworkMap: Record<string, string> = {
```
zacharydenton, line 11: Can we implement this as a function that takes in the aforementioned enum of wallet networks and returns an enum of valid explorer networks (i.e. `keyof typeof networks` - you will need to remove the `Record<string, string>` cast from `networks` for this to work - you should remove it from `walletNetworkMap` as well), with assertNever in the default case for exhaustiveness checking? This way if we ever add more wallet networks we will be required to update the mapping. This would replace the`getWalletNetworkFromNetworkMap` function later on in the PR.

src/context/wallet/context.tsx
```diff
@@ -4,6 +4,7 @@ export interface walletContext {
   isInstalled: boolean;
   isConnected: boolean;
   isAccountSet: boolean;
+  walletNetwork: string;
```
zacharydenton, line 7: This should be updated to the enum of wallet networks.

src/context/wallet/useWallet.tsx
```diff
@@ -28,27 +30,40 @@ export function useWallet() {
   useEffect(() => {
     // add this check to support older wallet versions
     if (isUpdatedVersion()) {
-      window.aptos.on("accountChanged", (account: any) => {
+      window.aptos?.on?.("accountChanged", (account: any) => {
```
zacharydenton, line 33: If window.aptos is not the object we expect it to be, should any of the code in this file be running?

src/GlobalState.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react";
-import {networks, defaultNetworkName} from "./constants";
+import {networks, defaultNetworkName, ExplorerNetworks} from "./constants";
 
-const selected_network = safeGetSelectedNetworkName();
+const selected_network = safeGetSelectedNetworkName() as ExplorerNetworks;
 
 function safeGetSelectedNetworkName(): string {
```
zacharydenton, line 6: Change this function to return `ExplorerNetwork` instead of `string`.

src/api/hooks/useSubmitTransaction.ts
```diff
@@ -1,5 +1,6 @@
 import {TxnBuilderTypes} from "aptos";
 import {useState} from "react";
+import {walleExplorertNetworkMap} from "../../constants";
```
zacharydenton, line 3: typo

src/constants.tsx
```diff
@@ -1,27 +1,41 @@
...
   devnet: devnetUrl,
   test: "https://rosetta.aptosdev.com/",
-  //governance: "https://ait2.aptosdev.com/", // TODO - use the proposals url
 };
 
-export const walletNetworkMap: Record<string, string> = {
-  Devnet: "devnet",
-  Localhost: "local",
-  Testnet: "test",
+export const walleExplorertNetworkMap = (
```
zacharydenton, line 12: Does this function need to be here or can it be in wallet-specific code?

src/constants.tsx
```diff
@@ -1,27 +1,41 @@
...
+    case "Testnet":
+      return "test";
+    default:
+      return assertNever(walletNetwork);
+  }
 };
 
+export type ExplorerNetworks = keyof typeof networks;
+
+export type WalletNetworks = "Devnet" | "Localhost" | "Testnet";
```
zacharydenton, line 29: Can this be in wallet-specific code?

src/constants.tsx
```diff
@@ -1,27 +1,41 @@
...
+
 // Remove trailing slashes
-for (let key of Object.keys(networks)) {
-  if (networks[key].endsWith("/")) {
-    networks[key] = networks[key].slice(0, -1);
+for (let value of Object.values(networks)) {
+  if (value.endsWith("/")) {
+    value = value.slice(0, -1);
```
zacharydenton, line 34: This looks like a bug, it's just creating new strings (not modifying the ones in the networks object).

src/constants.tsx
```diff
@@ -1,27 +1,41 @@
...
+    case "Localhost":
+      return "local";
+    case "Testnet":
+      return "test";
+    default:
+      return assertNever(walletNetwork);
+  }
 };
 
+export type ExplorerNetworks = keyof typeof networks;
```
zacharydenton, line 27: For consistency with the rest of the code, it should probably be called `NetworkName`.

src/context/wallet/context.tsx
```diff
@@ -1,10 +1,11 @@
 import {createContext, useContext} from "react";
+import {WalletNetworks} from "../../constants";
 
 export interface walletContext {
   isInstalled: boolean;
   isConnected: boolean;
   isAccountSet: boolean;
-  walletNetwork: string;
+  walletNetwork: WalletNetworks;
```
zacharydenton, line 8: Does it make sense to limit the scope of the wallet-specific network names (i.e. change this to `NetworkName` (the explorer network name, and getWalletNetwork() could also return NetworkName))?

msavir: I thought about it but it might be a bit confusing to set the `walletNetwork` (i.e Localhost/Devnet/Testnet) as an `explorerNetwork` (i.e local/devnet/test) where in fact they are different, it might makes more sense to have it as it really is (i.e Localhost/Devnet/Testnet), and when we need (for now for the validation) we can use the map we just created

src/GlobalState.tsx
```diff
@@ -1,21 +1,21 @@
...
   let selected_network = localStorage.getItem("selected_network");
   if (selected_network) {
     selected_network = selected_network.toLowerCase();
     if (selected_network in networks) {
-      return selected_network;
+      return selected_network as NetworkName;
     }
   }
-  return defaultNetworkName.toLowerCase();
+  return defaultNetworkName.toLowerCase() as NetworkName;
```
msavir, line 14: ts complains if I dont use cast also here and on line 11

zacharydenton (approved):
Thank you for adding types here!

PR #85, opened 2022-08-11
[Governance] Transaction response snackbar

zihan-aptos:
## In this PR:
* Modularized similar snackbar logic in the proposal page vote buttons and the stake page and created `TransactionResponseSnackbar`.
* Fix inaccurate snackbar response. We used to show success when there is a transaction hash returned. This is not correct because transaction hash means the transaction is submitted. The submitted transaction status could be either success/fail. Only successful transactions should be popping the GREEN snackbar alert. 
* Add transaction link in the snackbar.

https://user-images.githubusercontent.com/109111707/184064621-074cb602-6b91-4bca-8456-0767ab00a2ff.mov

## Next Steps
* Add loading indicator
* Auto update the Results. remove "Refresh" button.


msavir:
nice work! One thought is, the transaction success/failure response can take a while to come back and we might want to show a loader icon to the user to indicate something is happening

zihan-aptos:
> 

Good idea. I'll add it in a separate PR.

src/pages/Governance/components/TransactionResponseSnackbar.tsx
```diff
@@ -0,0 +1,222 @@
...
+  );
+}
+
+type SuccessSnackbarProps = {
+  transactionHash: string;
+  refreshOnSuccess?: boolean;
+  onCloseSnackbar: () => void;
+};
+
+function SuccessSnackbar({
```
msavir, line 61: might be nice to have success and failure response in 2 separate files for maintainability and readability. they can all be imported and exported from this file though

src/api/wallet.ts
```diff
@@ -66,15 +67,16 @@ export const signAndSubmitTransaction = async (
       transactionPayload,
     );
     if ("hash" in response) {
+      await client.waitForTransaction(response["hash"]);
```
msavir, line 70: I wonder if that something that needs to be happen on the wallet side, and we only should get the success/failure response?
@kent-white any thoughts?

zihan-aptos: Agree. This is a walk-around for now. I think eventually we want to have this from the wallet. Maybe something like `window.aptops.waitForTransaction()`.

src/api/hooks/useGetTransaction.ts
```diff
@@ -0,0 +1,16 @@
+import {Types} from "aptos";
+import {useQuery} from "react-query";
+import {getTransaction} from "../../api";
+import {ResponseError} from "../../api/client";
+import {useGlobalState} from "../../GlobalState";
+
+export function useGetTransaction(txnHashOrVersion: string) {
+  const [state, _setState] = useGlobalState();
+
+  const {data, status} = useQuery<Types.Transaction, ResponseError>(
```
zacharydenton, line 10: Why is the rest of the query result discarded?

src/pages/Governance/components/snackbar/SuccessSnackbar.tsx
```diff
@@ -0,0 +1,73 @@
...
+          Succeeded with transaction {""}
+          <Link
+            component={RRD.Link}
+            to={`/txn/${transactionHash}`}
+            color="inherit"
+            target="_blank"
+          >
+            {transactionHash}
+          </Link>
+          {refreshOnSuccess === true &&
```
zacharydenton, line 67: Why do we need to do a full page reload? Can't we just rerun the relevant queries? The explorer is relatively slow on initial page load.

src/pages/Governance/components/snackbar/TransactionResponseSnackbar.tsx
```diff
@@ -0,0 +1,83 @@
...
+}: TransactionResponseSnackbarProps) {
+  if (transactionResponse == null) {
+    return null;
+  }
+
+  if (!transactionResponse.transactionSubmitted) {
+    return (
+      <ErrorSnackbar
+        errorMessage={
+          (transactionResponse as TransactionResponseOnError).message
```
zacharydenton, line 47: Remove the type cast. Typescript should be able to infer that it's `TransactionResponseOnError` based on the `if (!transactionResponse.transactionSubmitted)` conditional.

src/pages/Governance/components/snackbar/TransactionResponseSnackbar.tsx
```diff
@@ -0,0 +1,83 @@
...
+        }
+        onCloseSnackbar={onCloseSnackbar}
+      />
+    );
+  }
+
+  const transactionHash = (
+    transactionResponse as TransactionResponseOnSubmission
```
zacharydenton, line 55: Remove the type cast

src/pages/Governance/components/snackbar/TransactionResponseSnackbar.tsx
```diff
@@ -0,0 +1,83 @@
...
+  ).transactionHash;
+  const {data, status} = useGetTransaction(transactionHash);
+
+  if (status !== "success") {
+    return null;
+  }
+
+  const isTransactionSuccess =
+    (data as Types.UserTransaction)?.success === true;
```
zacharydenton, line 64: Remove the type cast

src/pages/Governance/components/snackbar/TransactionResponseSnackbar.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  return (
+    <>
+      {isTransactionSuccess && (
+        <SuccessSnackbar
+          transactionHash={transactionHash as string}
```
zacharydenton, line 70: Remove the type cast

src/api/hooks/useGetTransaction.ts
zihan-aptos: For the snackbar response we only need `data` and `status`. We can probably add more in the future when they're needed.

src/pages/Governance/components/snackbar/SuccessSnackbar.tsx
zihan-aptos: Good point. It was a MVP design that at least prompt the users to check for the updated result. I'll add a TODO and refresh the "Results" part only.

src/pages/Governance/components/snackbar/TransactionResponseSnackbar.tsx
zihan-aptos: Oh nice! Didn't know that.

src/api/hooks/useGetTransaction.ts
msavir: it might worth returning the full result so then other components (https://github.com/aptos-labs/explorer/blob/main/src/pages/layout/Search.tsx#L256 , https://github.com/aptos-labs/explorer/blob/main/src/pages/Transactions/Transactions.tsx#L133) can use this hook. then each component can use whatever data it needs

zihan-aptos: made the change

msavir (approved):

PR #86, opened 2022-08-11
Add ait3 to networks

msavir:


zacharydenton:
Looks good but we should probably hold off on this change until the network is up and running.

PR #92, opened 2022-08-15
Upgrade aptos dep

msavir:
upgrade aptos dependency version


src/api/index.ts
```diff
@@ -30,48 +31,67 @@ export async function getAccountTransactions(
...
 
 export function getTransaction(
-  requestParameters: {txnHashOrVersion: string},
+  requestParameters: {txnHashOrVersion: string | bigint},
   nodeUrl: string,
 ): Promise<Types.Transaction> {
-  const client = new AptosClient(nodeUrl);
   const {txnHashOrVersion} = requestParameters;
-  return withResponseError(client.getTransaction(txnHashOrVersion));
+  if (isHex(txnHashOrVersion as string)) {
```
msavir, line 45: is this the correct format to use to check a txn hash?
cc @CapCap

src/pages/utils.tsx
```diff
@@ -133,3 +133,7 @@ export function isValidAccountAddress(accountAddr: string): boolean {
   // can start with that and see if any fails to parsing
   return /^(0[xX])?[a-fA-F0-9]{1,64}$/.test(accountAddr);
 }
+export function isHex(text: string) {
+  // if it's hex, and is <= (64 + 2 for 0x) char long
+  return text.startsWith("0x") && text.length <= 66;
```
zacharydenton, line 138: This seems similar to the regex on line 134 except this doesn't check that the chars are hex digits. Can you reuse the existing function?

src/api/index.ts
```diff
@@ -1,27 +1,28 @@
 import {AptosClient, Types} from "aptos";
+import {isHex} from "../pages/utils";
 import {withResponseError} from "./client";
 
 export async function getTransactions(
-  requestParameters: {start?: number; limit?: number},
+  requestParameters: {start?: bigint; limit?: number},
```
zacharydenton, line 6: JS numbers can safely represent integers up to 2^53 - 1 (roughly 9 quadrillion). I assume the SDK type changed, but we can keep our code the same without having to worry about BigInt by keeping the `getTransactions` signature the same and performing the number -> BigInt conversion inside it (maybe with an assertion that `0 <= start <= Number.MAX_SAFE_INTEGER`).

src/api/index.ts
```diff
@@ -1,27 +1,28 @@
...
+): Promise<Types.Transaction[]> {
   const client = new AptosClient(nodeUrl);
   const transactions = await withResponseError(
     client.getTransactions(requestParameters),
   );
 
   // Sort in descending order
   transactions.sort((a, b) =>
-    parseInt(a.version) < parseInt(b.version) ? 1 : -1,
+    parseInt((a as any).version) < parseInt((b as any).version) ? 1 : -1,
```
zacharydenton, line 16: Please do not cast values to `any`: it disables type checking and eventually leads to runtime errors. A better option here would be to narrow the type (e.g. with `'version' in a`). It's often helpful to hover over values in your editor to see what the inferred type is.

zacharydenton (commented):
If you bumped the version, there should be changes to `package.json` and/or `yarn.lock` - but I don't see them in this PR.

src/pages/utils.tsx
msavir: I used the same check it used before here https://github.com/aptos-labs/explorer/pull/92/files#diff-971be23a4da4e8d5a19915e6824596394b7b2b2b091130a93a7f3d38d8888a4cL234

Is tnx hex and account address have the same format?

zihan-aptos (commented):
I'm testing this PR on my local network. I'm seeing "Proposal Not Found" when clicking on one of the proposals in https://deploy-preview-92--aptos-explorer.netlify.app/proposals?network=local.

msavir:
> I'm testing this PR on my local network. I'm seeing "Proposal Not Found" when clicking on one of the proposals in https://deploy-preview-92--aptos-explorer.netlify.app/proposals?network=local.

I believe it is a cross error, I am getting it on previous PRs also

src/api/index.ts
```diff
@@ -1,27 +1,36 @@
...
 
 export async function getTransactions(
   requestParameters: {start?: number; limit?: number},
   nodeUrl: string,
-): Promise<Types.OnChainTransaction[]> {
+): Promise<Types.Transaction[]> {
   const client = new AptosClient(nodeUrl);
+  const {start, limit} = requestParameters;
+  let bigStart;
+  if (start && 0 <= start && start <= Number.MAX_SAFE_INTEGER) {
```
zacharydenton, line 12: I think this should be an assertion; if a number is passed in that's out of range, it's probably a programmer error and thus should blow up to notify the programmer (rather than silently omitting the param).

src/api/index.ts
```diff
@@ -1,27 +1,36 @@
...
+  }
   const transactions = await withResponseError(
-    client.getTransactions(requestParameters),
+    client.getTransactions({start: bigStart, limit}),
   );
 
   // Sort in descending order
   transactions.sort((a, b) =>
-    parseInt(a.version) < parseInt(b.version) ? 1 : -1,
+    ("version" in a && parseInt(a.version)) <
```
zacharydenton, line 21: If the transaction doesn't have a version property, that means it's a pending transaction (and thus it's expected version will be higher than any existing versions). You could consider the version to be Infinity for this case. It would be best to extract the sorting logic to a separate function to reuse later on in this file.

src/pages/Transactions/Transaction.tsx
```diff
@@ -293,7 +296,6 @@ function renderUserTransaction(transaction: Types.UserTransaction) {
           {renderRow("Gas Used:", renderGas(transaction.gas_used))}
           {renderRow("Max Gas:", renderGas(transaction.max_gas_amount))}
           {renderRow("Gas Unit Price:", renderGas(transaction.gas_unit_price))}
-          {renderRow("Gas Currency:", transaction.gas_currency_code)}
```
zacharydenton, line 296: Did it get renamed or is it meant to be removed entirely?

src/pages/Transactions/Transaction.tsx
```diff
@@ -366,7 +369,7 @@ function renderPendingTransaction(transaction: Types.PendingTransaction) {
 
 // TODO: Update type to StateCheckpointTransaction when it's added to the SDK.
```
zacharydenton, line 370: Remove this comment since the type is now in the SDK.

src/pages/Transactions/Transaction.tsx
```diff
@@ -382,8 +385,6 @@ function renderStateCheckpointTransaction(
           {renderRow("Version:", transaction.version)}
           {renderRow("Hash:", transaction.hash)}
           {renderRow("Status:", renderSuccess(transaction.success))}
-          {"proposer" in transaction &&
```
zacharydenton, line 385: Did it get renamed or is it meant to be removed entirely?

src/pages/Transactions/Transaction.tsx
msavir: it has been removed https://github.com/aptos-labs/aptos-core/pull/2902

src/api/index.ts
msavir: I think it goes out of scope, I will remove the `0 <= start && start <= Number.MAX_SAFE_INTEGER` for now and add a task to handle it later.

src/pages/Transactions/Transaction.tsx
msavir: was removed

zihan-aptos (approved):
Feedbacks have been addressed. AC to unblock.

PR #93, opened 2022-08-15
update explorer to prepare for internal testing

sherry-x:
- add sherrynet fullnode to explorer to make internal testing easier


msavir (approved):
looks good, but once merged it would be visible to all users, does that make sense?

sherry-x:
@msavir Yes that should be okay. would be nice to eventually have ability to gate the access but for now this is totally fine.

PR #95, opened 2022-08-15
[Governance] add loading spinner for transactions

zihan-aptos:
https://user-images.githubusercontent.com/109111707/184724362-a9991a7e-2406-4114-970c-26b6c45db1c1.mov

### DIFF
--- a/src/pages/Governance/Stake/Index.tsx
+++ b/src/pages/Governance/Stake/Index.tsx
@@ -1,4 +1,4 @@
-import React, {useEffect, useState} from "react";
+import React, {useEffect} from "react";
 import {Grid, Button, FormControl, Tooltip} from "@mui/material";
 import {Header} from "../components/Header";
 import {useWalletContext} from "../../../context/wallet/context";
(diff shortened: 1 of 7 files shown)


msavir:
with this change, is the voting option buttons and the submit button on the stake page are disabled also while we wait for the transaction result? (i.e users can keep clicking those buttons while they are waiting)

zihan-aptos:
> 

Good question. They cannot click on anything while waiting.

src/pages/Governance/Stake/Index.tsx
```diff
@@ -40,6 +42,7 @@ export function StakePage() {
     const isVoterAddrValid = validateVoterAddressInput();
 
     if (isStakingAmountValid && isOperatorAddrValid && isVoterAddrValid) {
+      setLoadingModalIsOpen(true);
```
msavir, line 45: would it make more sense to expose those functions/state from the `useSubmitTransaction` file? so it would live in one file?

msavir (approved):
nice

PR #98, opened 2022-08-16
Stake min max from chain config

msavir:
Pull the minimum/maximum staking amount from chain and validate user input on initializing staking pool

image

### DIFF
--- a/src/api/hooks/useAmountInput.tsx
+++ b/src/api/hooks/useAmountInput.tsx
@@ -1,16 +1,40 @@
 import React, {useState, useEffect} from "react";
 import AmountTextField from "../../components/AmountTextField";
+import {useGetAccountResources} from "./useGetAccountResources";
 
-// TODO - this will likely be very dependent per network
-// pull from on chain config 0x1::stake::ValidatorSetConfiguration
-function isValidAmount(amount: string): boolean {
+interface StakingConfigData {
+  minimum_stake: string;
+  maximum_stake: string;
+}
+
+function isValidAmount(
+  amount: string,
+  minimumAmount: number,
+  maximumAmount: number,
+): boolean {
   const amountNum = parseInt(amount);
-  return amountNum > 0;
+  return amountNum <= maximumAmount && amountNum >= minimumAmount;
 }
 
 const useAmountInput = () => {
   const [amount, setAmount] = useState<string>("");
   const [amountIsValid, setAmountIsValid] = useState<boolean>(true);
+  const [minimumAmount, setMinimumAmount] = useState<number>(0);
+  const [maximumAmount, setMaximumAmount] = useState<number>(0);
+
+  const accountResourcesResult = useGetAccountResources("0x1");
+
+  const stakingConfig = accountResourcesResult.data?.find(
+    (resource) => resource.type === "0x1::staking_config::StakingConfig",
+  );
+
+  useEffect(() => {
+    if (!stakingConfig) return;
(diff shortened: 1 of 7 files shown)


src/api/hooks/useAmountInput.tsx
```diff
@@ -1,16 +1,40 @@
...
+  amount: string,
+  minimumAmount: number,
+  maximumAmount: number,
+): boolean {
   const amountNum = parseInt(amount);
-  return amountNum > 0;
+  return amountNum <= maximumAmount && amountNum >= minimumAmount;
 }
 
 const useAmountInput = () => {
```
zihan-aptos, line 19: With the staking amount range logic added, this is now more of a `useStakingAmountInput`. Although this hook meant to be the general amount input hook to serve all kinds of amount. I don't have strong opinion on wether we should keep it that way though. It's okay if we want to make it a staking amount hook, maybe just change the name to reflect the current logic.

src/api/hooks/useAmountInput.tsx
```diff
@@ -1,16 +1,40 @@
...
+  const stakingConfig = accountResourcesResult.data?.find(
+    (resource) => resource.type === "0x1::staking_config::StakingConfig",
+  );
+
+  useEffect(() => {
+    if (!stakingConfig) return;
+    const {maximum_stake, minimum_stake} =
+      stakingConfig.data as StakingConfigData;
+    setMinimumAmount(minimum_stake ? parseInt(minimum_stake) : 0);
+    setMaximumAmount(maximum_stake ? parseInt(maximum_stake) : 0);
```
zihan-aptos, line 36: I'm not sure about the default min/max amounts here. If there's no `minimum_stake`, should we allow staking 0? Doesn't seem to be meaningful. If there's no `maximum_stake`, shouldn't we just allow any amount?

src/api/hooks/useAmountInput.tsx
```diff
@@ -1,16 +1,40 @@
...
   const amountNum = parseInt(amount);
-  return amountNum > 0;
+  return amountNum <= maximumAmount && amountNum >= minimumAmount;
 }
 
 const useAmountInput = () => {
   const [amount, setAmount] = useState<string>("");
   const [amountIsValid, setAmountIsValid] = useState<boolean>(true);
+  const [minimumAmount, setMinimumAmount] = useState<number>(0);
+  const [maximumAmount, setMaximumAmount] = useState<number>(0);
```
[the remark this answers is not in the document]
msavir: we need to wait for the response from `useGetAccountResources` to come back, on first load it in undefined and then need to update those values once we have a response. This is the use of useState (and useEffect)

src/api/hooks/useAmountInput.tsx
msavir: yes I was thinking on the option we dont get minimum_stake/maximum_stake back from chain - but I guess this case shouldn't happen, so we can assume we always have those values, will fix it

src/api/hooks/useAmountInput.tsx
msavir: Agree. let me work on it

msavir: I think it makes more sense to have it as a `stake amount input` instead of a global one for now, I started moving the validation logic into a staking related component and ended up with a plain TextField material ui component, also because staking page is the only place we use the amount input I think we can have it as a staking related file for now

zihan-aptos (approved):

PR #102, opened 2022-08-18
[Governance] add execution status to proposals table

zihan-aptos:
* make execution status helper functions and reuse them in the proposal header and proposals table
* add execution status to proposals table
Screen Shot 2022-08-18 at 12 21 18 PM

### DIFF
--- a/src/pages/Governance/utils.tsx
+++ b/src/pages/Governance/utils.tsx
@@ -1,10 +1,11 @@
 import React from "react";
 import {ensureMillisecondTimestamp} from "../utils";
-import {Proposal, ProposalVotingState} from "./Types";
+import {Proposal, ProposalVotingState, ProposalExecutionState} from "./Types";
 import CheckCircleOutlinedIcon from "@mui/icons-material/CheckCircleOutlined";
 import ErrorOutlineOutlinedIcon from "@mui/icons-material/ErrorOutlineOutlined";
 import PendingOutlinedIcon from "@mui/icons-material/PendingOutlined";
 import HighlightOffIcon from "@mui/icons-material/HighlightOff";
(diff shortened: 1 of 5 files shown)


msavir:
Great idea! Although imo the ui looks a bit busy, what do you think on having an info icon (maybe https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Ainfo%3A) and when hover it would say "waiting to be executed"?

src/pages/Governance/utils.tsx
```diff
@@ -122,3 +123,25 @@ export function renderStatusIcon(
       return assertNever(proposalState);
   }
 }
+
+export function getExecutionStatusColor(isResolved: boolean): string {
+  return isResolved ? primaryColor : warningColor;
+}
+
+export function renderExecutionStatusIcon(isResolved: boolean): JSX.Element {
```
zacharydenton, line 131: For stuff like this, just make it a component. `export function ExecutionStatusIcon({isResolved}: {isResolved: boolean}) { ... }`

zacharydenton (approved):
LGTM

The "Waiting to be executed" copy is kind of funny. Maybe just "Pending"

zihan-aptos:
> 

I'd keep it as consistent as possible as the proposal page, meaning that I'd prefer show two status. Although I agree that the UI can be more clean. Two solutions I'm thinking:
1. As Zach mentioned below, the phrase "Waiting to be executed" is a bit redundant. If we can come up with a shorter but accurate phrase, the UI would look better. 
2. I'm thinking if we should just have 5 status for voting+execution instead having two set of status. They can be: Pending, Failed, Rejected, Passed (implying waiting for execution), Executed. 
It'll be a good discussion to have with the team. Let me bring this to a broader group.

PR #103, opened 2022-08-18
[Governance] replace proposal creation time with voting expiration time

zihan-aptos:
Screen Shot 2022-08-18 at 4 04 25 PM
Screen Shot 2022-08-18 at 4 03 54 PM

 
Screen Shot 2022-08-19 at 10 00 56 AM

### DIFF
--- a/src/pages/Governance/Proposal/Header.tsx
+++ b/src/pages/Governance/Proposal/Header.tsx
@@ -67,14 +67,12 @@ function ExecutionStatusComponent({
 function TimeRemainingComponent({
   proposal,
   isOnMobile,
-  isVotingClosed,
 }: {
   proposal: Proposal;
   isOnMobile: boolean;
-  isVotingClosed: boolean;
 }) {
   // TODO: show close time if it's closed
-  if (isVotingClosed) {
+  if (isVotingClosed(proposal)) {
     return null;
   }
 
@@ -96,11 +94,9 @@ function TimeRemainingComponent({
 function ProposerAndTimeComponent({
   proposal,
   isOnMobile,
-  isVotingClosed,
 }: {
   proposal: Proposal;
   isOnMobile: boolean;
-  isVotingClosed: boolean;
 }) {
   const proposerComponent = (
     <Stack direction="row" spacing={1}>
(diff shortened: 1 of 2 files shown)


msavir:
I would suggest to have ` start date` and `end date` (same as https://snapshot.org/#/ens.eth/proposal/0x5c96e490f3e28d8269e8fc7e929491fb8fa5e4bd04d3379f0c4f4bb1a42dc23e here). 
On the table we can have the `end date` (although that might not be necessary also)

zihan-aptos:
> I would suggest to have ` start date` and `end date` (same as https://snapshot.org/#/ens.eth/proposal/0x5c96e490f3e28d8269e8fc7e929491fb8fa5e4bd04d3379f0c4f4bb1a42dc23e here). On the table we can have the `end date` (although that might not be necessary also)

Added "Start Date" and "End Date" columns for the proposals table. For the proposal page, with the current design, we don't have a good place to include two fields "Start Date" and "End Date", so I added "Voting Period" to represent both of them.

msavir:
> ave a good place to include two fields "Start Date" and "End Date", so I added "Voting Period" to represent both of them.

looks better! I am a bit worried on how the "voting period" would look like on mobile view - what do you think on having the `end date` for now (on the proposal page - as this is why we started with this change) and discuss the other changes with the UI/UX folks?

zacharydenton (approved):

msavir (approved):
add a comment, other than that lgtm

zihan-aptos:
> > ave a good place to include two fields "Start Date" and "End Date", so I added "Voting Period" to represent both of them.
> 
> looks better! I am a bit worried on how the "voting period" would look like on mobile view - what do you think on having the `end date` for now (on the proposal page - as this is why we started with this change) and discuss the other changes with the UI/UX folks?

I added screenshot for "voting period" on mobile view. I wouldn't worry too much about the mobile view right now since we don't expect a lot of mobile users for AIT3. I would prefer showing voting period than an end date only because it makes more sense for the proposal page to have all the information that the proposal table row has. But let me discuss it with the team and revisit it later for better UI.

PR #106, opened 2022-08-19
handle new proposal metadata property

msavir:
We were missing executed proposals from the UI because the metadata property is gone from the proposal object once the proposal has been executed. 
A new property was introduced to keep the metadata alive https://github.com/aptos-labs/aptos-core/pull/3140/files 

We now get the metadata values as a utf8 string encoded as hex, we need to convert it back to string.

### DIFF
--- a/src/pages/Governance/Types.ts
+++ b/src/pages/Governance/Types.ts
@@ -4,10 +4,7 @@ export type Proposal = {
   proposer: string;
   creation_time_secs: string;
   execution_content: {
-    vec: Array<{
-      metadata_location: string;
-      metadata_hash: string;
-    }>;
+    vec: [];
   };
   execution_hash: string;
   min_vote_threshold: number;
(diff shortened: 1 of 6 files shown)


src/pages/Governance/Types.ts
```diff
@@ -4,10 +4,7 @@ export type Proposal = {
   proposer: string;
   creation_time_secs: string;
   execution_content: {
```
zihan-aptos, line 6: If we don't need it anymore, just remove this field?

zihan-aptos (approved):

src/pages/Governance/Types.ts
msavir: It is not deprecated on the move module https://github.com/aptos-labs/aptos-core/pull/3140/files#diff-5b53920d43083fa179b5920517df200f09135ca238f81d8971a45efd7ee5936eR60 so thought we might want/need to have it around

PR #111, opened 2022-08-22
fix to match with Petra's custom network feature

msavir:
Petra introduced a "custom networks" where the uncustomized networks are Devnet and Testnet (and in the future probably also "Mainnet"). Explorer does a check (by network name) if the dApp network is same as Petra network before submitting a transaction, to make it work with Petra's custom url feature, it should do this check only for the uncustomized networks.

### DIFF
--- a/src/constants.tsx
+++ b/src/constants.tsx
@@ -3,8 +3,8 @@ export const devnetUrl =
 
 export const networks = {
   local: "http://localhost:8080",
-  devnet: devnetUrl,
-  test: "https://rosetta.aptosdev.com/",
+  Devnet: devnetUrl,
+  Testnet: "https://rosetta.aptosdev.com/",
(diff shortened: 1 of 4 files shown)


src/constants.tsx
```diff
@@ -3,8 +3,8 @@ export const devnetUrl =
 
 export const networks = {
   local: "http://localhost:8080",
-  devnet: devnetUrl,
-  test: "https://rosetta.aptosdev.com/",
+  Devnet: devnetUrl,
+  Testnet: "https://rosetta.aptosdev.com/",
```
msavir, line 7: @CapCap what is `rosetta` used for? On Petra, `Testnet` points to the future ait3 network.

zihan-aptos: I believe it's a partner test network. Coinbase?

zihan-aptos (approved):

PR #112, opened 2022-08-22
Use Petra 1.6 version

msavir:
Updates to use Petra 0.1.6 version

1. Pass transactions as JSON object to Petra
2. Display Petra's transactions errors

### DIFF
--- a/src/pages/Governance/hooks/useSubmitStake.ts
+++ b/src/pages/Governance/hooks/useSubmitStake.ts
@@ -1,4 +1,4 @@
-import {BCS, TxnBuilderTypes} from "aptos";
+import {Types} from "aptos";
 import useSubmitTransaction from "../../../api/hooks/useSubmitTransaction";
 
 const useSubmitStake = () => {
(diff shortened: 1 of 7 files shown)


src/pages/Governance/hooks/useSubmitStake.ts
```diff
@@ -10,22 +10,16 @@ const useSubmitStake = () => {
   } = useSubmitTransaction();
 
   async function submitStake(
-    stakingAmount: number,
+    stakingAmount: string,
```
msavir, line 13: should be string for now, once Petra move to use ABI interface we should be able to send a number
https://aptos-org.slack.com/archives/C03N83P7QUC/p1660945235254379

src/pages/Governance/hooks/useSubmitVote.ts
```diff
@@ -10,26 +10,16 @@ const useSubmitVote = () => {
   } = useSubmitTransaction();
 
   async function submitVote(
-    proposalId: number,
+    proposalId: string,
```
msavir, line 13: should be string for now, once Petra move to use ABI interface we should be able to send a number
https://aptos-org.slack.com/archives/C03N83P7QUC/p1660945235254379

CapCap (approved):

PR #114, opened 2022-08-23
Governance sub menu

msavir:
Desktop

image

Mobile

image

image

### DIFF
--- a/src/components/Nav.tsx
+++ b/src/components/Nav.tsx
@@ -70,7 +70,7 @@ export default function Nav() {
           Proposals
         </MenuItem>
         <MenuItem onClick={() => handleCloseAndNavigate("/proposals/stake")}>
-          Staking
+          Stake
         </MenuItem>
       </Menu>
     </Box>
--- a/src/components/Nav.tsx
+++ b/src/components/Nav.tsx
@@ -69,8 +69,8 @@ export default function Nav() {
         <MenuItem onClick={() => handleCloseAndNavigate("/proposals")}>
           Proposals
         </MenuItem>
-        <MenuItem onClick={() => handleCloseAndNavigate("/proposals/stake")}>
-          Stake
+        <MenuItem onClick={() => handleCloseAndNavigate("/proposals/staking")}>
+          Staking
         </MenuItem>
       </Menu>
     </Box>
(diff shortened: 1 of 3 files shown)


src/components/Nav.tsx
```diff
@@ -38,6 +57,22 @@ export default function Nav() {
...
+        anchorEl={governanceMenuEl}
+        TransitionComponent={Fade}
+        aria-controls={open ? "fade-menu" : undefined}
+        aria-haspopup="true"
+        aria-expanded={open ? "true" : undefined}
+      >
+        <MenuItem onClick={() => handleCloseAndNavigate("/proposals")}>
+          Proposals
+        </MenuItem>
+        <MenuItem onClick={() => handleCloseAndNavigate("/proposals/stake")}>
```
zihan-aptos, line 72: Have we decided on the name "Staking"? If so, I would change the route to `/proposals/staking`.

zihan-aptos (commented):
Looks good! I'll leave it to @moonclavedev

moonclavedev:
> Desktop
> 
> image
> 
> Mobile
> 
> image

Overall it's good but we will have to do some work on managing the text hierarchy here as it's a little awkward on mobile - this is more obvious because we don't have a top-level toggle with the arrow on this like we do on the marketing site. 

Maybe a quick fix for now would be to put a down arrow on the right as if it were already expanded? 

image

msavir:
> > Desktop
> > image
> > Mobile
> > image
> 
> Overall it's good but we will have to do some work on managing the text hierarchy here as it's a little awkward on mobile - this is more obvious because we don't have a top-level toggle with the arrow on this like we do on the marketing site.
> 
> Maybe a quick fix for now would be to put an arrow on the right as if it were already expanded? This does break the rules and users might be confused as to why they can't toggle it however.
> 
> image

@moonclavedev great idea!

something like that?

image

image

moonclavedev:
Also, can we change "Stake" to "Staking"? @CapCap

moonclavedev:
> > > Desktop
> > > image
> > > Mobile
> > > image
> > 
> > 
> > Overall it's good but we will have to do some work on managing the text hierarchy here as it's a little awkward on mobile - this is more obvious because we don't have a top-level toggle with the arrow on this like we do on the marketing site.
> > Maybe a quick fix for now would be to put an arrow on the right as if it were already expanded? This does break the rules and users might be confused as to why they can't toggle it however.
> > image
> 
> @moonclavedev great idea!
> 
> something like that?
> 
> image
> 
> image

Yep!

moonclavedev (approved):

PR #119, opened 2022-08-24
use state to show stake pool, voter, operator changes on the ui

msavir:
After users initialize staking pool OR change operator OR change voter, we want to show the new app state (without the need for them to refresh the page)

https://user-images.githubusercontent.com/29798064/186471548-806724a3-5615-4e7f-b0ce-434951c319ea.mov


zihan-aptos (approved):
Code looks good! Can we add a demo video to show the change?

PR #120, opened 2022-08-24
increase lockup button

msavir:
Owner can increase lockup through the UI

image

### DIFF
--- a/src/pages/Governance/Stake/Edit.tsx
+++ b/src/pages/Governance/Stake/Edit.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import {Grid} from "@mui/material";
+import {Divider, Grid} from "@mui/material";
 import {Types} from "aptos";
 import {EditOperator} from "./components/EditOperator";
 import {EditVoter} from "./components/EditVoter";
@@ -29,8 +29,8 @@ export function Edit({stakePool, isWalletConnected}: EditProps) {
             isWalletConnected={isWalletConnected}
           />
         </Grid>
-        <Grid item xs={12} sm={12}>
-          <hr />
+        <Grid item xs={12}>
+          <Divider variant="dotted" orientation="horizontal" />
         </Grid>
         <Grid item xs={12} sm={12}>
           <EditVoter
@@ -38,8 +38,8 @@ export function Edit({stakePool, isWalletConnected}: EditProps) {
             isWalletConnected={isWalletConnected}
           />
         </Grid>
-        <Grid item xs={12} sm={12}>
-          <hr />
+        <Grid item xs={12}>
+          <Divider variant="dotted" orientation="horizontal" />
         </Grid>
         <Grid item xs={12} sm={12}>
           <IncreaseLockup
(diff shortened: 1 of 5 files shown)


zihan-aptos:
Would an owner want to increase lockup before staking? It might be better to make increase lockup a separate page so that owners can do it before staking too. Also increasing lockup is a different flow/feature/function than staking and it makes more sense to have it own page.

msavir:
> Would an owner want to increase lockup before staking? It might be better to make increase lockup a separate page so that owners can do it before staking too. Also increasing lockup is a different flow/feature/function than staking and it makes more sense to have it own page.

this page is **after** initialize staking. Only accounts that have staking pool see this page. It makes sense to have this page as the "owner dashboard" where they can make changes to their staking pool imo

msavir:
> Would an owner want to increase lockup before staking? It might be better to make increase lockup a separate page so that owners can do it before staking too. Also increasing lockup is a different flow/feature/function than staking and it makes more sense to have it own page.

I am not sure they can increase lockup without having a StakePool resource....?

zihan-aptos:
> > Would an owner want to increase lockup before staking? It might be better to make increase lockup a separate page so that owners can do it before staking too. Also increasing lockup is a different flow/feature/function than staking and it makes more sense to have it own page.
> 
> I am not sure they can increase lockup without having a StakePool resource....? Having the increase lockup on this page makes sense as this page is kind of an "owner dashboard" where they can make changes to their staking pool imo

@movekevin please let us know if owners can/need increase lockup

src/pages/Governance/Stake/Edit.tsx
```diff
@@ -12,34 +13,41 @@ type EditProps = {
...
+    <>
+      <Grid container spacing={6}>
+        <Grid item xs={12} sm={12}>
+          <EditOperator
+            operatorAddress={operator_address}
+            isWalletConnected={isWalletConnected}
+          />
+        </Grid>
+        <Grid item xs={12} sm={12}>
+          <hr />
```
zihan-aptos, line 33: Can we use `<Divider variant="dotted" orientation="horizontal" />`

src/pages/Governance/Stake/components/IncreaseLockup.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <TransactionResponseSnackbar
+        transactionResponse={transactionResponse}
+        onCloseSnackbar={onCloseSnackbar}
+      />
+      <LoadingModal open={transactionInProcess} />
+      <Grid>
+        <Grid container spacing={4}>
+          <></>
+          <Grid item xs={12} sm={8}>
+            Locked until: {renderTimestamp(lockedUntilSec)}
```
zihan-aptos, line 52:
```suggestion
            LOCKED UNTIL: {renderTimestamp(lockedUntilSec)}
```

zihan-aptos (approved):

src/pages/Governance/Stake/components/IncreaseLockup.tsx
```diff
@@ -0,0 +1,69 @@
...
+  );
+
+  return (
+    <>
+      <TransactionResponseSnackbar
+        transactionResponse={transactionResponse}
+        onCloseSnackbar={onCloseSnackbar}
+      />
+      <LoadingModal open={transactionInProcess} />
+      <Grid>
```
zihan-aptos, line 48: can we set alignItems to `center` for the text and the button?

movekevin (approved):

PR #121, opened 2022-08-24
check if account has already voted on a proposal

msavir:
When the connected account (on Petra) enters a proposal page we want to let them know whether they already voted on that proposal.

When proposal is still open to votes
Screen Shot 2022-08-24 at 4 04 23 PM

When proposal is close to votes
Screen Shot 2022-08-24 at 4 04 29 PM

When not voted
image

### DIFF
--- a/src/pages/Governance/Proposal/hooks/useAccountHasVoted.tsx
+++ b/src/pages/Governance/Proposal/hooks/useAccountHasVoted.tsx
@@ -1,0 +1,71 @@
+import {useEffect, useState} from "react";
+
+import {getTableItem} from "../../../../api";
+import {useGetAccountResource} from "../../../../api/hooks/useGetAccountResource";
+import {useWalletContext} from "../../../../context/wallet/context";
+import {GlobalState, useGlobalState} from "../../../../GlobalState";
+
+const fetchTableItem = async (
+  pool_address: string,
+  proposal_id: string,
+  handle: string,
+  state: GlobalState,
+): Promise<boolean> => {
+  const tableItemRequest = {
+    key_type: "0x1::aptos_governance::RecordKey",
+    value_type: "bool",
+    key: {
+      stake_pool: pool_address,
+      proposal_id,
+    },
+  };
+
+  await getTableItem(
+    {tableHandle: handle, data: tableItemRequest},
+    state.network_value,
+  );
+
+  return true;
+};
+
+interface VotingRecords {
+  votes: {
+    handle: string;
+  };
+}
+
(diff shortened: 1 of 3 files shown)


src/pages/Governance/Proposal/hooks/useAccountHasVoted.tsx
```diff
@@ -0,0 +1,71 @@
...
+  pool_address: string;
+}
+
+export function useAccountHasVoted(proposalId: string): boolean {
+  const [state, _setState] = useGlobalState();
+  const {accountAddress} = useWalletContext();
+  const [hasVoted, setHadVoted] = useState<boolean>(false);
+
+  const {accountResource: ownerCapabilityResource} = useGetAccountResource(
+    accountAddress || "0x1",
```
movekevin, line 47: Why default to 0x1 here? 0x1 will never hold OwnerCapability

src/pages/Governance/Proposal/hooks/useAccountHasVoted.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    const votingResource = votingRecordResource.data as VotingRecords;
+    const {handle} = votingResource.votes;
+
+    const ownerCapability = ownerCapabilityResource.data as OwnerCapability;
+    const {pool_address} = ownerCapability;
+
+    fetchTableItem(pool_address, proposalId, handle, state)
+      .then(setHadVoted)
+      .catch(() => setHadVoted(false)); // fetchTableItem returns a 404 error if the item is not in the table
```
movekevin, line 67: What if there's another error such as a 500?

zihan-aptos:
Thanks for the swift work! It should solve the pain point for a lot of users. I have some idea on the UI design. If you don't mind I can try to put up a PR after you land this one.

zihan-aptos (approved):

src/pages/Governance/Proposal/hooks/useAccountHasVoted.tsx
msavir: its something I am working on, accountAddress might be null and the request function expects a string type (nut null)

src/pages/Governance/Proposal/hooks/useAccountHasVoted.tsx
msavir: Good point! in that case it is a server error and we should also set it to false - I think the bigger question is if there is a server error in (any) of the requests, what should we display? I will address it in a later PR

PR #122, opened 2022-08-24
[Governance] add execution time

zihan-aptos:
Screen Shot 2022-08-24 at 4 06 32 PM
Screen Shot 2022-08-24 at 4 21 08 PM

### DIFF
--- a/src/pages/Governance/Proposal/Header.tsx
+++ b/src/pages/Governance/Proposal/Header.tsx
@@ -23,14 +23,40 @@ function TitleComponent({proposal}: {proposal: Proposal}): JSX.Element {
 
 function StatusComponent({proposal}: {proposal: Proposal}): JSX.Element {
   return (
-    <ProposalStatusTooltip>
-      <Box sx={{display: "flex", alignItems: "center", gap: 1}}>
-        <StatusIcon status={proposal.status} />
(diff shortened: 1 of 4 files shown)


src/pages/Governance/Proposal/Header.tsx
```diff
@@ -23,14 +23,40 @@ function TitleComponent({proposal}: {proposal: Proposal}): JSX.Element {
...
+      </ProposalStatusTooltip>
+      {ExecutionTimeComponent(proposal)}
+    </Stack>
+  );
+}
+
+function ExecutionTimeComponent(proposal: Proposal) {
+  // TODO: remove proposal.resolution_time_secs null check after AIT3,
+  // as we shouldn't have null resolution_time_secs by then
+  if (!(proposal.is_resolved && proposal.resolution_time_secs)) {
```
movekevin, line 46: Why not if (!proposal.is_resolved || !proposal.resolution_time_secs) ? So if the proposal has not been resolved or the resolution time secs is not available. This makes more sense to me

movekevin (approved):

src/pages/Governance/Proposal/Header.tsx
zihan-aptos: I think they're essentially the same. But I'll change it for better readability.

PR #124, opened 2022-08-25
[Governance] Show vote participation

zihan-aptos:
This PR added a new result bar called "Participation" to display the following info:
* The current votes percentage of total voting power
* The minimum votes percentage of total voting power
* Detailed info in the tooltip on hover
Screen Shot 2022-08-24 at 6 57 56 PM
Screen Shot 2022-08-24 at 6 57 21 PM


src/pages/Governance/hooks/useGetCoinSupplyLimit.tsx
```diff
@@ -0,0 +1,43 @@
...
+  };
+  symbol: string;
+}
+
+export function useGetCoinSupplyLimit(): string | null {
+  const accountResourcesResult = useGetAccountResources("0x1");
+
+  if (!accountResourcesResult.data) return null;
+
+  const coinInfo = accountResourcesResult.data.find(
```
msavir, line 32: can utilize `useGetAccountResource` hook https://github.com/aptos-labs/explorer/blob/main/src/api/hooks/useGetAccountResource.ts

src/pages/Governance/hooks/useGetCoinSupplyLimit.tsx
```diff
@@ -0,0 +1,43 @@
...
+        integer: {
+          vec: [];
+        };
+      },
+    ];
+  };
+  symbol: string;
+}
+
+export function useGetCoinSupplyLimit(): string | null {
```
msavir, line 27: is it possible that this function return `null`? i.e that `0x1` will not have the `0x1::coin::CoinInfo<0x1::aptos_coin::AptosCoin>` resource? if not, we might want to utilize `useEffect` (and potentially `useState`) to update the returned data

src/pages/Governance/hooks/useGetCoinSupplyLimit.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  const coinInfo = accountResourcesResult.data.find(
+    (resource) =>
+      resource.type === "0x1::coin::CoinInfo<0x1::aptos_coin::AptosCoin>",
+  );
+
+  if (!coinInfo || !coinInfo.data) return null;
+
+  const coinInfoData: CoinInfo = coinInfo.data as CoinInfo;
+  const supplyLimit = coinInfoData?.supply?.vec[0]?.aggregator?.vec[0]?.limit;
```
msavir, line 40: tough =)

msavir (approved):

src/pages/Governance/Proposal/card/ParticipationBar.tsx
```diff
@@ -0,0 +1,160 @@
...
+}
+
+function TooltipContent(
+  currentVotes: number,
+  minVotes: number,
+  totalSupply: number,
+): JSX.Element {
+  const currentPercentage = ((currentVotes / totalSupply) * 100).toFixed(0);
+  const minPercentage = ((minVotes / totalSupply) * 100).toFixed(0);
+  const enoughParticipation = currentVotes > minVotes;
```
movekevin, line 30: Should be >=

src/pages/Governance/Proposal/card/ParticipationBar.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+type ParticipationBarProps = {
+  proposal: Proposal;
+};
+
+export default function ParticipationBar({proposal}: ParticipationBarProps) {
+  const theme = useTheme();
+
+  const supplyLimit = useGetCoinSupplyLimit();
+  if (!supplyLimit) {
```
movekevin, line 66: nit: I'd prefer an explicit undefined check: if supplyLimit === undefined

src/pages/Governance/Proposal/card/ParticipationBar.tsx
```diff
@@ -0,0 +1,160 @@
...
+    return null;
+  }
+
+  const totalSupply = parseInt(supplyLimit);
+  const currentVotes = getTotalVotes(proposal);
+  const minVotes = proposal.min_vote_threshold;
+
+  const currentPercentage = (currentVotes / totalSupply) * 100;
+  const enoughParticipation = currentVotes > minVotes;
```
movekevin, line 75: Again this should be >=

src/pages/Governance/Proposal/card/ParticipationBar.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const part1Percentage = enoughParticipation
```
movekevin, line 76: What are all these part1/part2/part3 percentages?

src/pages/Governance/hooks/useGetCoinSupplyLimit.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const accountResourcesResult = useGetAccountResources("0x1");
+
+  if (!accountResourcesResult.data) return null;
+
+  const coinInfo = accountResourcesResult.data.find(
+    (resource) =>
+      resource.type === "0x1::coin::CoinInfo<0x1::aptos_coin::AptosCoin>",
+  );
+
+  if (!coinInfo || !coinInfo.data) return null;
```
movekevin, line 37: nit: Use explicit === undefined checks instead

src/pages/Governance/hooks/useGetCoinSupplyLimit.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  const coinInfoData: CoinInfo = coinInfo.data as CoinInfo;
```
movekevin, line 39: nit: Here and in some other places: Why is the explicit type declaration (: CoinInfo) needed? TS can infer this from the context.

src/pages/Governance/Proposal/card/ParticipationBar.tsx
zihan-aptos: I divided the bar in 3 parts, left, middle and right. The left is `Min(currentVotes, minVotes)`; the middle is either `|minVotes-currentVotes|`; and the right is `(totalSupply - Max(currentVotes, minVotes))`.

src/pages/Governance/hooks/useGetCoinSupplyLimit.tsx
zihan-aptos: Actually TS reports error without explicit type declaration. :(

src/pages/Governance/hooks/useGetCoinSupplyLimit.tsx
zihan-aptos: I think right now we don't have much to worry about it since unlike `useGetProposal`, here we don't have any async functions to wait.

msavir: correct, but the return type is `string | null` and I wonder if there is a real option that coin supply will return `null` or we just use the `null` return type to support the condition checking

zihan-aptos: Oh I got it. So here just use `null` as condition checking.

PR #125, opened 2022-08-25
[Governance] show voted

zihan-aptos:
This PR iterate upon #121 to:
* Design change: make "VOTED" part as a section inside the card.
* Logic change: hide CastVoteSection component when the current account is voted.

|Not voted yet|Voted|
|---|---|
|Screen Shot 2022-08-25 at 12 15 04 PM|Screen Shot 2022-08-25 at 12 15 35 PM|


src/pages/Governance/Proposal/card/YourVoteSection.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import Section from "./Section";
+import HowToVoteIcon from "@mui/icons-material/HowToVote";
+import {Stack, Typography} from "@mui/material";
+
+// TODO: add the specific vote (yes/no) after indexer is ready
+export default function YourVoteSection(): JSX.Element {
```
zacharydenton, line 7: This component would be more flexible if it accepted `{...rest}` props and then passed them into the Section component (this way you could add a margin to it, for example). It's a good idea to consider any time you're making a component.

zacharydenton (approved):

msavir (approved):

PR #126, opened 2022-08-25
[Explorer] fix hash button link

zihan-aptos:
A few days ago I added a link for HashButon so that clicking on the expanded hash button will bring us to the account page for that hash. The bug was, the link is set to account page for all hashes. Clicking on a transaction hash button also brings us to account page which is wrong. This PR fixes the bug by adding `type` property for the component `HashButton` and redirecting to the correct page on click.

### DIFF
--- a/src/components/HashButton.tsx
+++ b/src/components/HashButton.tsx
@@ -21,24 +21,27 @@ export enum HashType {
   OTHERS = "others",
 }
 
+function getHashLinkStr(hash: string, type: HashType): string {
+  switch (type) {
+    case HashType.ACCOUNT:
+      return `/account/${hash}`;
+    case HashType.TRANSACTION:
+      return `/txn/${hash}`;
+    case HashType.OTHERS:
+      return "";
+    default:
+      return assertNever(type);
+  }
+}
+
 function HashLink(hash: string, type: HashType): JSX.Element {
   switch (type) {
     case HashType.ACCOUNT:
-      return (
-        <Link
-          component={RRD.Link}
-          to={`/account/${hash}`}
-          target="_blank"
-          color="inherit"
-        >
-          {hash}
-        </Link>
-      );
     case HashType.TRANSACTION:
       return (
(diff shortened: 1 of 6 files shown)


src/components/HashButton.tsx
```diff
@@ -13,12 +13,51 @@ import {grey} from "../themes/colors/aptosColorPalette";
...
+import {assertNever} from "../utils";
+
+export enum HashType {
+  ACCOUNT = "account",
+  TRANSACTION = "transaction",
+  OTHERS = "others",
+}
+
+function HashLink(hash: string, type: HashType): JSX.Element {
+  switch (type) {
```
msavir, line 25: why not just passing the url? (i.e `account/txn`) that way dont need to create new element (that are pretty much the same) for each "type"? 

```suggestion
 function HashLink(hash,url){
 return (
 <Link
          component={RRD.Link}
          to={`/${url}/${hash}`}
          target="_blank"
          color="inherit"
        >
          {hash}
        </Link>
 )
 }
```

src/components/HashButton.tsx
```diff
@@ -13,12 +13,51 @@ import {grey} from "../themes/colors/aptosColorPalette";
...
+        <Link
+          component={RRD.Link}
+          to={`/txn/${hash}`}
+          target="_blank"
+          color="inherit"
+        >
+          {hash}
+        </Link>
+      );
+    case HashType.OTHERS:
```
msavir, line 48: is it being used somewhere?

src/components/HashButton.tsx
msavir: or if you prefer keeping a type, can hold a `type=>url` map and apply the url to the Link component based on the type

src/components/HashButton.tsx
zihan-aptos: Not yet but I figured it's good to have it

msavir (approved):

PR #131, opened 2022-08-29
[Governance] staking page wallet connection hot fix

zihan-aptos:
To address bug report https://aptos-org.slack.com/archives/C03SS1MMX3K/p1661805542133359.
|Before|After|
|---|---|
|Screen Shot 2022-08-29 at 4 09 56 PM|Screen Shot 2022-08-29 at 4 10 31 PM|

This is a hot fix to show the staking page even when the connected account is not in the right network. 

Ideally, we should prevent users connect their accounts on the wrong network in the wallet extension. This will take some decent effort and some time to wait for Chrome store review. So here we implement this hot fix in the Governance UI as a temporary solution. 

The hot fix is safe because when users try to submit the staking transaction with wrong account-network relationship, the transaction will fail anyway.

### DIFF
--- a/src/pages/Governance/Stake/components/CreateOrEdit.tsx
+++ b/src/pages/Governance/Stake/components/CreateOrEdit.tsx
@@ -1,5 +1,6 @@
 import React, {useEffect, useState} from "react";
 import {useGetAccountResource} from "../../../../api/hooks/useGetAccountResource";
+import LoadingModal from "../../components/LoadingModal";
 import {Create} from "../Create";
 import {Edit} from "../Edit";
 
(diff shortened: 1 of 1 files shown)


src/pages/Governance/Stake/components/CreateOrEdit.tsx
```diff
@@ -31,13 +27,9 @@ export function CreateOrEdit({
     refetch();
   };
 
-  if (isLoading) return <LoadingModal open={isLoading} />;
```
zacharydenton, line 34: Does the loading indicator need to be removed?

zacharydenton (approved):

PR #138, opened 2022-09-01
Add basic NHC UI

banool:


zihan-aptos:
If we'll only have this NHC page during AIT3, can we take it out from the nav bar? So that only people with the specific URL can visit it. This could reduce the UX inconsistency for the Explorer as a whole. What do you think @CapCap @moonclavedev

src/pages/NodeChecker/Client.tsx
```diff
@@ -0,0 +1,58 @@
+import {EvaluationSummary, NodeCheckerClient} from "aptos-node-checker-client";
+import {GlobalState} from "../../GlobalState";
+
+export const DEFAULT_NHC_INSTANCE =
```
CapCap, line 4: What happens when someone switches networks?

CapCap (approved):

zihan-aptos:
Thanks for updating the nab bar!

zihan-aptos (approved):

src/pages/NodeChecker/Client.tsx
banool: If they select "local" it tries to use a local NHC. If they select anything else, it uses the prod NHC (this default instance value).

PR #145, opened 2022-09-02
Format APT balances correctly

CapCap:
see tests for cases :-)

### DIFF
--- a/src/pages/Account/AccountInfo/Balance.tsx
+++ b/src/pages/Account/AccountInfo/Balance.tsx
@@ -25,17 +25,17 @@ export function getFormattedBalanceStr(balance: string, decimals?: number): stri
   }
 
   // Otherwise, insert decimal point at len - decimals
-  const left_side = BigInt(balance.slice(0, len - decimals)).toLocaleString("en-US");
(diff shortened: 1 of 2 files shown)


gregnazario (approved):

src/pages/Account/AccountInfo/Balance.tsx
```diff
@@ -2,14 +2,40 @@ import React from "react";
...
+  const len = balance.length;
+  decimals = decimals || APTOS_DECIMALS;
+
+  // If length is less than decimals, pad with 0s to decimals length and return
+  if (len <= decimals) {
+    return "0." + "0".repeat(decimals - len) + balance;
+  }
+
+  // Otherwise, insert decimal point at len - decimals
+  const left_side = BigInt(balance.slice(0, len - decimals)).toLocaleString("en-US");
```
zacharydenton, line 28: Don't use snake case in JS

zacharydenton (approved):

src/pages/Account/AccountInfo/Balance.tsx
CapCap: but what if the the world is wrong 😭

PR #146, opened 2022-09-09
[Explorer] address search bar feedback

zihan-aptos:
In this PR:
* address search bar feedback from the community: https://discord.com/channels/945856774056083548/946142976814497853/1012763801076117514
  * only show txn/account results when the input text is a valid txnHash/txnVersion/accountAddress
  * for transactions, if the user searches with txnHash, show txnHash in the result; if the user searches with version, show version in the result. (right now it's always version even searching with txnHash).
* clean up code for the search bar

https://user-images.githubusercontent.com/109111707/189247516-5fbe9817-0382-4fd3-a40f-0d5fab2d6879.mov


src/pages/layout/Search/Index.tsx
```diff
@@ -0,0 +1,132 @@
+import React, {useState, useEffect, useMemo} from "react";
+import {throttle} from "lodash";
+import {Autocomplete, AutocompleteInputChangeReason} from "@mui/material";
+import SearchTransaction from "./SearchTransaction";
+import SearchAccount from "./SearchAccount";
+import SearchInput from "./SearchInput";
+import ResultPaper from "./ResultPaper";
+import SearchResultNotFound from "./SearchResultNotFound";
+
+const HEX_REGEXP = /^(0x)?[0-9a-fA-F]+$/;
```
msavir, line 10: there is this function, https://github.com/aptos-labs/explorer/blob/main/src/pages/utils.tsx#L143 , it is a different checking but the naming can confused (one of them need to be changed...?)

src/pages/layout/Search/SearchAccount.tsx
```diff
@@ -0,0 +1,23 @@
...
+import ResultLink from "./ResultLink";
+
+type SearchAccountProps = {
+  address: string;
+};
+
+export default function SearchAccount({address}: SearchAccountProps) {
+  const {isLoading, error, data} = useGetAccountResources(address);
+
+  if (isLoading || error || !data) {
```
msavir, line 12: we can show a spinner when isLoading https://github.com/aptos-labs/explorer/blob/main/src/pages/Governance/Stake/components/CreateOrEdit.tsx#L36

src/pages/layout/Search/SearchInput.tsx
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+import {
+  FormControl,
+  InputAdornment,
+  SvgIcon,
+  SvgIconProps,
+  TextField,
+} from "@mui/material";
+
+function SearchIcon(props: SvgIconProps) {
```
msavir, line 10: why not using a search icon...? https://mui.com/material-ui/material-icons/?query=search&selected=Search

src/pages/layout/Search/Index.tsx
```diff
@@ -0,0 +1,132 @@
...
+  // hide autocomplete dropdown until text has been entered
+  const [open, setOpen] = useState(false);
+
+  const handleInputChange = (
+    event: any,
+    newInputValue: React.SetStateAction<string>,
+    reason: AutocompleteInputChangeReason,
+  ) => {
+    if (newInputValue.length === 0) {
+      if (open) setOpen(false);
```
msavir, line 77: nit: `setOpen(!open)`

src/pages/layout/Search/Index.tsx
```diff
@@ -0,0 +1,132 @@
...
+    } else {
+      if (!open) setOpen(true);
+    }
+    if (event && event.type === "blur") {
+      setInputValue("");
+    } else if (reason !== "reset") {
```
msavir, line 83: do we need to handle the `reset` reason case also...?

src/pages/layout/Search/SearchTransaction.tsx
```diff
@@ -0,0 +1,25 @@
...
+import {useGetTransaction} from "../../../api/hooks/useGetTransaction";
+
+type SearchTransactionProps = {
+  txnHashOrVersion: string;
+};
+
+export default function SearchTransaction({
+  txnHashOrVersion,
+}: SearchTransactionProps) {
+  const {isLoading, isError, data} = useGetTransaction(txnHashOrVersion);
```
msavir, line 12: `useGetTransaction` doesnt return isLoading/isError https://github.com/aptos-labs/explorer/blob/main/src/api/hooks/useGetTransaction.ts - might be good to modify it to return it (see https://github.com/aptos-labs/explorer/blob/main/src/api/hooks/useGetAccountResource.ts)

src/pages/layout/Search/Index.tsx
zihan-aptos: `isHex` is not exact accurate because it returns `false` for inputs like `1` which is a valid address and version. I have a task in my TODO list to unify this https://linear.app/aptoslabs/issue/WP-169/[explorer]-0x-address-format. Will revisit it later with this task.

src/pages/layout/Search/SearchAccount.tsx
zihan-aptos: I'm not sure if it's the best idea here even though showing spinner is a common practice in similar scenarios. The reason being: `isLoading` will be true for a while (5-10 seconds) when the txnHash or address is invalid.

src/pages/layout/Search/SearchInput.tsx
zihan-aptos: Good question, I'm just reusing what we've been using. Let me make the change.

src/pages/layout/Search/Index.tsx
zihan-aptos: I would actually prefer `setOpen (true)` or `setOpen (false)` over `setOpen(!open)`. It's easier to read and less confusing when someone tries to make change to the code later.

src/pages/layout/Search/Index.tsx
zihan-aptos: TBH I copied this part directly from Max's old code and didn't look very deep into it. I'd trust what has been working well before I figure out otherwise.

src/pages/layout/Search/SearchTransaction.tsx
zihan-aptos: Sounds reasonable. Although I'm curious why we don't want to just return the full result. This way whoever uses this hook has the flexibility to select what fields are useful. Also see discussion here: https://github.com/aptos-labs/explorer/pull/85#discussion_r943964465.

msavir: can do that too =) . just make sure it works also with [useGetAccountResource](https://github.com/aptos-labs/explorer/blob/main/src/api/hooks/useGetAccountResource.ts), as it has a different implementation than the others. also, need to fix it in here also https://github.com/aptos-labs/explorer/pull/146#discussion_r966524748

msavir: now, both `useGetTransaction` and `useGetAccountResources` doesnt return `isLoading/isError/error` or anything else than `data`

zihan-aptos: Sorry I'm not completely following. Do we want to make change to the return or this is looking good for now? Regardless, I feel that this issue worth a separate PR as this one is already pretty big and self-contained. Might worth it to address it in separate one.

msavir (approved):

PR #150, opened 2022-09-13
[Explorer] add json card

zihan-aptos:
https://user-images.githubusercontent.com/109111707/190016113-ac635431-d040-4697-a3ef-a9b3ffec674c.mov

### DIFF
--- a/src/components/JsonCard.tsx
+++ b/src/components/JsonCard.tsx
@@ -1,0 +1,70 @@
+import React, {useState} from "react";
+import {Box, useTheme} from "@mui/material";
+import KeyboardArrowUpIcon from "@mui/icons-material/KeyboardArrowUp";
+import KeyboardArrowDownIcon from "@mui/icons-material/KeyboardArrowDown";
+
+const TEXT_COLOR_LIGHT = "#0EA5E9";
+const TEXT_COLOR_DARK = "#83CCED";
+const BACKGROUND_COLOR = "rgba(14,165,233,0.1)";
+
+type JsonCardProps = {
+  data: any;
+};
+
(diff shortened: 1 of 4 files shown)


src/components/JsonCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <pre
+          style={{
+            margin: 0,
+            fontFamily: theme.typography.fontFamily,
+            fontWeight: theme.typography.fontWeightRegular,
+            fontSize: 13,
+            overflowWrap: "break-word",
+          }}
+        >
+          {JSON.stringify(data || null, null, 2)}
```
msavir, line 53: do we want to display something if there is no `data`?

src/components/JsonCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+import {Box, useTheme} from "@mui/material";
+import KeyboardArrowUpIcon from "@mui/icons-material/KeyboardArrowUp";
+import KeyboardArrowDownIcon from "@mui/icons-material/KeyboardArrowDown";
+
+const TEXT_COLOR_LIGHT = "#0EA5E9";
+const TEXT_COLOR_DARK = "#83CCED";
+const BACKGROUND_COLOR = "rgba(14,165,233,0.1)";
+
+type JsonCardProps = {
+  data: any;
```
msavir, line 11: should we use a specific type here?

msavir (approved):

zacharydenton (approved):

src/components/JsonCard.tsx
zihan-aptos: It's not quite possible because this component will be reused by `event.data` whose type is `any`.

PR #158, opened 2022-09-15
[Explorer] new txn page: add tooltips

zihan-aptos:
In this PR:
* reordered fileds in txn Overview tabs
* added tooltips for fields to explain what it is and provide useful links to learn more
* code clean up
Screen Shot 2022-09-15 at 3 16 42 PM

### DIFF
--- a/src/components/IndividualPageContent/LearnMoreTooltip.tsx
+++ b/src/components/IndividualPageContent/LearnMoreTooltip.tsx
@@ -1,0 +1,57 @@
+import React from "react";
+import HelpOutlineOutlinedIcon from "@mui/icons-material/HelpOutlineOutlined";
+import {Box, Link, Typography, Tooltip, useTheme} from "@mui/material";
+import {grey} from "../../themes/colors/aptosColorPalette";
+import {Stack} from "@mui/system";
+
+function TooltipBox({children}: {children?: React.ReactNode}) {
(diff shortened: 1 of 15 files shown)


src/components/IndividualPageContent/LearnMoreTooltip.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <Tooltip
+        title={
+          <Stack alignItems="flex-end">
+            {linkToText ? (
+              <Link alignSelf="flex-end" href={link} color="inherit">
+                {text}
+              </Link>
+            ) : (
+              <>
+                <Typography variant="inherit">{text}</Typography>
```
zacharydenton, line 37: Monospace doesn't seem like the write font choice for a tooltip.

zacharydenton (approved):

src/components/IndividualPageContent/LearnMoreTooltip.tsx
moonclavedev: Agreed. If we could specify Tooltips to utilize **apparat** that would be ideal. Understandably the default font config for Explorer is **lft-etica-mono**, so we should probably create a `Typography` variant with `font-family:"apparat";`

zihan-aptos: Working on it.

PR #162, opened 2022-09-20
[Explorer] set up Graphql and token page

zihan-aptos:
* created token page
* query token data from the indexer with graphql
Screen Shot 2022-09-20 at 12 28 21 PM

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -3,6 +3,12 @@ import ReactDOM from "react-dom";
 import {BrowserRouter} from "react-router-dom";
 import {QueryClient, QueryClientProvider} from "react-query";
 import ExplorerRoutes from "./ExplorerRoutes";
+import {
+  ApolloClient,
+  InMemoryCache,
+  ApolloProvider,
+  HttpLink,
+} from "@apollo/client";
 
 import * as Sentry from "@sentry/react";
 import {BrowserTracing} from "@sentry/tracing";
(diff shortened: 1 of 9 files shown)


src/index.tsx
```diff
@@ -32,15 +38,28 @@ declare global {
 
 const queryClient = new QueryClient();
 
+const graphqlClient = new ApolloClient({
+  link: new HttpLink({
+    uri: "https://profound-koala-32.hasura.app/v1/graphql",
```
zacharydenton, line 43: These details look like they should be environment variables.

src/pages/Token/Index.tsx
```diff
@@ -0,0 +1,68 @@
...
+import {gql, useQuery} from "@apollo/client";
+import {Stack, Grid} from "@mui/material";
+import React from "react";
+import HeaderSearch from "../layout/Search/Index";
+import TokenTitle from "./Title";
+import TokenTabs from "./Tabs";
+
+const TOKEN_QUERY = gql`
+  {
+    token_datas(where: {name_hash: {}, name: {_eq: "Aptos Zero: 7"}}) {
```
zacharydenton, line 11: Remove `name: {_eq: "Aptos Zero: 7"}`?

src/pages/Token/Index.tsx
```diff
@@ -0,0 +1,68 @@
...
+      name
```
zacharydenton, line 12: Does the UI require all of these fields? It is preferable to load only what is needed.

src/pages/Token/Index.tsx
zihan-aptos: We'll be querying the token_datas by token_id. Bowen is still working on adding the token_id so I query it by name as a temporary solution. The idea is to have some data to fill the token page. I'll change it once we have the token_id ready.

msavir (approved):

PR #168, opened 2022-09-22
[Explorer] update aptos client

zihan-aptos:
* update aptos client version from `1.3.7` to `1.3.14`
* apply changes to work with the new version

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "@types/lodash": "^4.14.179",
     "@types/react": "latest",
     "@types/react-dom": "latest",
-    "aptos": "^1.3.7",
+    "aptos": "^1.3.14",
     "aptos-node-checker-client": "0.0.2",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "@types/lodash": "^4.14.179",
     "@types/react": "latest",
     "@types/react-dom": "latest",
-    "aptos": "^1.3.7",
+    "aptos": "^1.3.14",
     "aptos-node-checker-client": "0.0.2",
     "date-fns": "^2.29.1",
     "graphql": "^16.6.0",
```
jjleng, line 20: shouldn't `typescript` and `prettier` be dev dependencies?

msavir (approved):

package.json
zihan-aptos: yes it in line 35 and line 24 if that's what you mean?

PR #173, opened 2022-09-26
[Explorer] block page more

zihan-aptos:
This PR added more fields to the block page Overview tab.
Screen Shot 2022-09-26 at 12 08 35 PM

### DIFF
--- a/src/pages/Block/Tabs/OverviewTab.tsx
+++ b/src/pages/Block/Tabs/OverviewTab.tsx
@@ -4,6 +4,8 @@ import React from "react";
 import ContentBox from "../../../components/IndividualPageContent/ContentBox";
 import ContentRow from "../../../components/IndividualPageContent/ContentRow";
 import TimestampValue from "../../../components/IndividualPageContent/ContentValue/TimestampValue";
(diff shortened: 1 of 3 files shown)


zacharydenton (approved):

src/pages/Block/Tabs/OverviewTab.tsx
```diff
@@ -20,20 +22,63 @@ function VersionValue({data}: {data: Types.Block}) {
...
+}
+
 type OverviewTabProps = {
   data: Types.Block;
 };
 
 export default function OverviewTab({data}: OverviewTabProps) {
+  const blockTxn: Types.Transaction | undefined = (
+    data.transactions ?? []
+  ).find((txn) => {
```
zacharydenton, line 59: (For some reason, this comment didn't get posted)

Instead of casting it to the type in the component, it would be better to create a type guard (i.e. `is` instead of `as`).

https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

zihan-aptos: Thanks for the tip! Applied changes in #174

PR #175, opened 2022-09-26
[Explorer] show owned tokens in account page

zihan-aptos:
Screen Shot 2022-09-26 at 3 40 41 PM

### DIFF
--- a/src/pages/Account/Tabs/TokensTab.tsx
+++ b/src/pages/Account/Tabs/TokensTab.tsx
@@ -1,0 +1,37 @@
+import React from "react";
+import {gql, useQuery} from "@apollo/client";
+import {TokensTable} from "../Components/TokensTable";
+
+const TOKENS_QUERY = gql`
+  query TokensData($owner_address: String) {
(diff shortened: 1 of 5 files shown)


src/pages/Account/Tabs/TokensTab.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import {gql, useQuery} from "@apollo/client";
+import {TokensTable} from "../Components/TokensTable";
+
+const TOKENS_QUERY = gql`
```
msavir, line 5: does this query give us all (currently) owned tokens? meaning it excludes ones they sold?

msavir (approved):

src/pages/Account/Tabs/TokensTab.tsx
zihan-aptos: It's a good question for @bowenyang007. My guess is that it excludes the sold ones.

bowenyang007: plz do amount > 0 lol

PR #179, opened 2022-09-27
[Explorer] nav bar: added new menu items

zihan-aptos:
Screen Shot 2022-09-27 at 12 40 26 PM
Screen Shot 2022-09-27 at 12 41 52 PM

### DIFF
--- a/src/components/Nav.tsx
+++ b/src/components/Nav.tsx
@@ -24,7 +24,7 @@ function NavButton({
       title={title}
       sx={{
         color: "inherit",
-        fontSize: {sm: ".875rem", md: "1rem"},
+        fontSize: "1rem",
       }}
     >
       {label}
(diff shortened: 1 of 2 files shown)


src/components/Nav.tsx
```diff
@@ -34,45 +61,55 @@ export default function Nav() {
...
+      />
+      {inDev && (
+        <>
+          <NavButton to="/blocks" title="View All Blocks" label="Blocks" />
+          <NavButton
+            to="/Validators"
+            title="View All Validators"
+            label="Validators"
+          />
+          <NavButton to="/tokens" title="View Top Tokens" label="Tokens" />
```
msavir, line 77: do we want to use the `MenuItem` component as we do with all the other items?
https://github.com/aptos-labs/explorer/pull/179/files#diff-7f2ef60bbec771e3e2a3e1e7162d91e481d7724cd76ea86a82a0b4ad45c3baefR102

msavir (approved):

src/components/Nav.tsx
zihan-aptos: We want them to be buttons in the header. `MenuItem` will be used later when we have sub menu for the buttons. For example: 
+ Tokens
    + Top Tokens
    + Top Collections

PR #181, opened 2022-09-28
[Explorer] transaction table code clean up

zihan-aptos:
Code clean up. No observable changes.


src/components/TransactionType.tsx
```diff
@@ -7,51 +7,70 @@ import MultipleStopRoundedIcon from "@mui/icons-material/MultipleStopRounded";
...
     case "state_checkpoint_transaction":
-      return <OutlinedFlagIcon fontSize="small" />;
+      return <OutlinedFlagIcon fontSize="small" color={color} />;
     default:
       throw `Unknown TransactionType:${type}`;
   }
 }
 
-type TransactionTypeProps = {
+type Props = {
```
msavir, line 56:
```suggestion
type TransactionType = {
```

src/components/TransactionType.tsx
```diff
@@ -7,51 +7,70 @@ import MultipleStopRoundedIcon from "@mui/icons-material/MultipleStopRounded";
 import UpdateRoundedIcon from "@mui/icons-material/UpdateRounded";
 import {grey} from "../themes/colors/aptosColorPalette";
 
+type Color =
```
msavir, line 10: ochh I also had it a couple of days ago, are we allowing all colors? we can restrict to only the color types we use in the app. from looking at the code here it seems like we only use "primary"?

msavir (approved):

PR #186, opened 2022-09-30
[Explorer] show user transactions only on home page

zihan-aptos:
Transactions type filtering has not been supported yet. Although we do want users to see user transactions only, at least in the home page. Therefore this PR is a walk around before we have transaction type filtering supported.
Screen Shot 2022-09-30 at 2 16 57 PM

### DIFF
--- a/src/api/hooks/useGetUserTransactionVersions.ts
+++ b/src/api/hooks/useGetUserTransactionVersions.ts
@@ -1,0 +1,43 @@
+import {gql, useQuery as useGraphqlQuery} from "@apollo/client";
+
+const USER_TRANSACTIONS_FROM_QUERY = gql`
+  query UserTransactions($limit: Int, $start_version: bigint) {
+    user_transactions(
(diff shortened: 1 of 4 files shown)


src/api/hooks/useGetUserTransactionVersions.ts
```diff
@@ -0,0 +1,43 @@
...
+  const {loading, error, data} = useGraphqlQuery(
+    startVersion ? USER_TRANSACTIONS_FROM_QUERY : USER_TRANSACTIONS_QUERY,
+    {variables: {limit: limit, start_version: startVersion}},
+  );
+
+  if (loading || error || !data) {
+    return [];
+  }
+
+  const versions: number[] = (data?.user_transactions ?? []).map(
```
msavir, line 36: do you need to check `data?.user_transactions ?? []` if you already check it on line 32?

msavir (approved):

PR #188, opened 2022-10-02
prepare for Premainnet

sherry-x:


zacharydenton (approved):
Looks like this also needs to be changed: https://github.com/aptos-labs/explorer/blob/8ffda6b5f53b18043dbd93fda14b53e27fd5e316/src/api/hooks/useGetCoinSupplyLimit.tsx#L56

cc @zihan-aptos

sherry-x:
I put the ait3 back for now and we can do another PR to clean it up

PR #190, opened 2022-10-03
Update Page Title logic, Metric Cards, Color Palette & Mobile Stacking

moonclavedev:
- Update Page Title Styling & Logic
- Bring updated Metric Card styling inline with current theme
- Updated color palette
- Address some mobile stacking issues

image

image

image

### DIFF
--- a/src/components/IndividualPageContent/ContentRow.tsx
+++ b/src/components/IndividualPageContent/ContentRow.tsx
@@ -1,5 +1,6 @@
 import React from "react";
-import {Grid, Box, Typography, Stack} from "@mui/material";
+import {Box, Typography, Stack} from "@mui/material";
+import Grid from "@mui/material/Unstable_Grid2";
 import {grey} from "../../themes/colors/aptosColorPalette";
 import EmptyValue from "./ContentValue/EmptyValue";
 
@@ -20,28 +21,34 @@ export default function ContentRow({
     <Box>
       <Grid
         container
-        rowSpacing={1}
+        rowSpacing={0.5}
         columnSpacing={4}
-        alignItems="center"
+        alignItems="start"
         key={i}
       >
-        <Grid item xs={12} md={3}>
-          <Stack direction="row">
-            {tooltip}
+        <Grid xs={12} sm={3}>
+          <Box>
             <Typography variant="body2" color={grey[450]}>
               {title}
+              <Box
+                component="span"
+                sx={{display: "inline", whiteSpace: "nowrap"}}
+              >
+                &nbsp;
+                <Box sx={{display: "inline-block"}}>{tooltip}</Box>
+              </Box>
             </Typography>
-          </Stack>
-        </Grid>
-        <Grid item xs={12} md={9}>
-          <Box
-            sx={{
-              fontSize: 13.5,
-              overflow: "auto",
-            }}
-          >
-            {value ? <Box>{value}</Box> : <EmptyValue />}
           </Box>
+        </Grid>
+        <Grid
+          xs={12}
+          sm={9}
+          sx={{
+            fontSize: "0.8rem",
+            overflow: "auto",
+          }}
+        >
+          {value ? <Box>{value}</Box> : <EmptyValue />}
         </Grid>
       </Grid>
     </Box>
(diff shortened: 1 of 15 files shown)


src/themes/colors/aptosColorPalette.ts
```diff
@@ -13,18 +13,18 @@ export const grey = {
...
+  300: "#3ffdff",
+  400: "#06f7f7",
+  500: "#00dadc",
+  600: "#00adb2",
+  700: "#00878d",
+  800: "#00686f",
+  900: "#06565b",
 };
 
 export const aptosColor = "#02858D";
```
zihan-aptos, line 30: can we remove this?

src/pages/Transaction/Title.tsx
```diff
@@ -16,23 +16,23 @@ export default function TransactionTitle({transaction}: TransactionTitleProps) {
...
       spacing={2}
       marginX={1}
       marginBottom={2}
     >
       {isOnMobile ? (
         <>
-          <Stack direction="row" alignItems="center" spacing={2}>
-            <Typography variant="h5">Transaction:</Typography>
+          <Stack direction="column" alignItems="center" spacing={2}>
+            <Typography variant="h3">Transaction</Typography>
```
zihan-aptos, line 28: I think we want to have the same title variant for all pages? I'll change for the rest pages in a follow up PR.

src/components/IndividualPageContent/ContentRow.tsx
```diff
@@ -33,15 +33,17 @@ export default function ContentRow({
...
+          item
+          md={9}
+          width={{xs: 1, md: 0.75}}
+          sx={{
+            fontSize: 13.5,
+            overflow: "auto",
+            marginLeft: 3,
+          }}
+        >
+          {value ? <Box>{value}</Box> : <EmptyValue />}
```
zihan-aptos, line 46: Is this expected? https://deploy-preview-190--aptos-explorer.netlify.app/txn/151759208?network=testnet&feature=dev. The labels and values are in separate lines on non-mobile too.

zihan-aptos (commented):
Thanks for the work!! This is beautiful. Just one question for `ContentRow` before shipping this PR.

src/components/IndividualPageContent/ContentRow.tsx
```diff
@@ -22,7 +22,7 @@ export default function ContentRow({
         container
         rowSpacing={1}
         columnSpacing={4}
-        alignItems="center"
+        alignItems="start"
```
zihan-aptos, line 25: I would make it as "center" because the values are not always a line of text. Sometimes it's a hash button, sometimes it's a json card. I feel it looks better when the label and value are center aligned with each other.

src/themes/colors/aptosColorPalette.ts
moonclavedev: I would leave for now as this provides all shades of our Primary color, which may be used in the future.

src/components/IndividualPageContent/ContentRow.tsx
```diff
@@ -20,29 +21,35 @@ export default function ContentRow({
...
+        <Grid xs={12} sm={3}>
+          <Box>
             <Typography variant="body2" color={grey[450]}>
               {title}
+              <Box
+                component="span"
+                sx={{display: "inline", whiteSpace: "nowrap"}}
+              >
+                &nbsp;
+                <Box sx={{display: "inline-block"}}>{tooltip}</Box>
```
zihan-aptos, line 38: Looks great! Thanks! One last thought: would it be even better to make the info icon center-aligned with the label?

zihan-aptos (approved):
Feel free

zihan-aptos (approved):

src/components/IndividualPageContent/ContentRow.tsx
zihan-aptos: @moonclavedev please feel free to land it. I can come back to this later. I'd love to have the changes landed for the dogfooding session/instruction I'll soon be sending out.

src/components/IndividualPageContent/ContentRow.tsx
moonclavedev: The problem with vertically centering here is if there is a large content area to the right, the label could potentially be pushed below the viewport. 

In terms of hierarchy the label / title should be presented before (or at the same time) as it's content.

PR #191, opened 2022-10-03
[Explorer] config graphql client by network

zihan-aptos:
* Right now the graphql client is set to testnet (hard-coded). This PR create a `GraphqlClientProvider` and config the graphql client by selected network.
* For networks that do not have graphql clients available, this PR hide components/pages that need to query data from graphql.


src/api/hooks/useGraphqlClient.tsx
```diff
@@ -0,0 +1,89 @@
...
+    case "testnet":
+      return true;
+    case "premainnet":
+      return false;
+    default:
+      return false;
+  }
+}
+
+function getGraphqlURI(networkName: NetworkName): string | undefined {
```
msavir, line 28: nit: could use this fun to also know if `getIsGraphqlClientSupportedFor`

msavir (approved):

PR #199, opened 2022-10-06
[Explorer] search on "Enter" key press

zihan-aptos:
https://user-images.githubusercontent.com/109111707/194185702-f50fb5a3-1840-4a39-9fc8-c43215191240.mov

### DIFF
--- a/src/pages/layout/Search/Index.tsx
+++ b/src/pages/layout/Search/Index.tsx
@@ -12,7 +12,6 @@ import {useGetInDevMode} from "../../../api/hooks/useGetInDevMode";
 import {getAccount, getTransaction} from "../../../api";
 import {useGlobalState} from "../../../GlobalState";
 import {useNavigate} from "react-router-dom";
-import {resolve} from "@sentry/utils";
 
 const HEX_REGEXP = /^(0x)?[0-9a-fA-F]+$/;
 
@@ -182,11 +181,15 @@ export default function HeaderSearch() {
       onInputChange={handleInputChange}
       onClose={() => setOpen(false)}
       renderInput={(params) => {
-        return <SearchInput {...params} onSubmitSearch={handleSubmitSearch} />;
+        return <SearchInput {...params} />;
       }}
       renderOption={(props, option) => {
         if (!option) return null;
         return <li {...props}>{option}</li>;
+      }}
+      onSubmit={(event) => {
+        handleSubmitSearch();
+        event.preventDefault();
       }}
     />
   );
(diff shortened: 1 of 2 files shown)


src/pages/layout/Search/Index.tsx
```diff
@@ -101,6 +106,24 @@ export default function HeaderSearch() {
     }
   };
 
+  const handleEnterPress = async () => {
+    try {
+      await getAccount({address: inputValue}, state.network_value);
```
zacharydenton, line 111: Can you start both requests simultaneously instead of sequentially? e.g. `const accountPromise = getAccount(); const transactionPromise = getTransaction(); ... await accountPromise; ... await transactionPromise;` This will reduce latency if it's a transaction.

src/pages/layout/Search/SearchInput.tsx
```diff
@@ -23,9 +35,14 @@ export default function SearchInput({...params}) {
             </InputAdornment>
           ),
         }}
-        placeholder="Search transactions"
-        helperText="Account Address / Txn Version & Hash / Block Height & Version"
+        placeholder="Search Explorer"
+        helperText="Account Address / Txn Hash or Version / Block Height or Version"
         fullWidth
+        onKeyPress={(event) => {
```
zacharydenton, line 41: I think it makes more sense to have it be in onSubmit on the form. This will work better if the form is submitted programmatically (i.e. it's more accessible).

src/pages/layout/Search/Index.tsx
```diff
@@ -101,6 +107,45 @@ export default function HeaderSearch() {
     }
   };
 
+  const handleSubmitSearch = async () => {
+    const accountPromise = getAccount(
+      {address: inputValue},
+      state.network_value,
+    )
+      .then(() => {
```
zacharydenton, line 115: You could use async / await as before, the request will start when you call the function.

src/pages/layout/Search/SearchInput.tsx
```diff
@@ -1,10 +1,28 @@
...
+  onSubmitSearch: () => void;
+}
+
+export default function SearchInput({
+  onSubmitSearch,
+  ...params
+}: SearchInputProps) {
   return (
-    <FormControl sx={{width: "100%"}}>
+    <form
```
zacharydenton, line 19: Isn't there a parent component that has a form? An input component should not be creating a form.

zihan-aptos: Unfortunately I couldn't find a parent component that has a form in MUI.

src/pages/layout/Search/Index.tsx
zihan-aptos: Sorry I'm not a JS expert. I'm a bit confused. I thought we wanted to do 2 queries simultaneously. Would you give me more instruction or code snippet?

PR #201, opened 2022-10-06
[Explorer] fix bugs reported from dogfooding

zihan-aptos:
Fixed:
* event key is deprecated and should not be shown
* The Devnet URL network parameter requires an uppercase "Devnet" value unlike the other networks.
* Transactions and Validators have "back" button but Blocks doesnt
* auto redirect to the new governance site

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -20,14 +20,6 @@ if (window.location.origin.includes("explorer.devnet.aptos.dev")) {
   window.location.replace(new_location);
 }
 
-if (window.location.origin.includes("explorer.aptoslabs.com/proposals")) {
-  const new_location = window.location.href.replace(
-    "explorer.aptoslabs.com/proposals",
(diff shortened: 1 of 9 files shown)


src/index.tsx
```diff
@@ -20,6 +20,14 @@ if (window.location.origin.includes("explorer.devnet.aptos.dev")) {
   window.location.replace(new_location);
 }
 
+if (window.location.origin.includes("explorer.aptoslabs.com/proposals")) {
```
msavir, line 23: do we need it? we will deprecate the governance pages, so there will be no way to enter those pages on the client side, and we already handle the redirect on the server side

msavir (approved):

src/index.tsx
zihan-aptos: oh right I forgot that we've handled it on the serverside

PR #202, opened 2022-10-07
[Explorer] fixing bugs from dogfooding

zihan-aptos:
* bolding the tab name visited while on it (Transactions, Block, Validators). 
* show "No Data" card message when no data on Transactions tab on account page (right now it shows an empty table)
* json card:
    * limit the max height of the expanded card
    * when card is expanded, only the collapse button is clickable, not the whole card
* validators page:
    * make validators table not clickable
    * table alignment
    * validators ordered in voting power

### DIFF
--- a/src/components/GeneralTableRow.tsx
+++ b/src/components/GeneralTableRow.tsx
@@ -5,30 +5,39 @@ import {grey} from "../themes/colors/aptosColorPalette";
 interface TableRowProps {
   children?: React.ReactNode;
   onClick: () => void;
+  clickDisabled?: boolean;
 }
 
-export default function GeneralTableRow({children, onClick}: TableRowProps) {
+export default function GeneralTableRow({
+  children,
+  onClick,
+  clickDisabled,
+}: TableRowProps) {
   const theme = useTheme();
   return (
     <TableRow
       onClick={onClick}
       sx={{
-        cursor: "pointer",
+        cursor: clickDisabled ? undefined : "pointer",
         userSelect: "none",
         backgroundColor: `${
           theme.palette.mode === "dark" ? grey[800] : grey[50]
         }`,
-        "&:hover:not(:active)": {
-          filter: `${
-            theme.palette.mode === "dark"
-              ? "brightness(0.9)"
-              : "brightness(0.99)"
-          }`,
-        },
-        "&:active": {
-          background: theme.palette.neutralShade.main,
-          transform: "translate(0,0.1rem)",
-        },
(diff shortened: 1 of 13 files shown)


src/components/GeneralTableRow.tsx
```diff
@@ -5,30 +5,39 @@ import {grey} from "../themes/colors/aptosColorPalette";
...
+  onClick,
+  clickDisabled,
+}: TableRowProps) {
   const theme = useTheme();
   return (
     <TableRow
       onClick={onClick}
       sx={{
-        cursor: "pointer",
+        cursor: clickDisabled ? undefined : "pointer",
```
zacharydenton, line 21: Does `clickDisabled` actually prevent clicking or does it just adjust the visual style? Might want to adjust the onClick callback to exit early if clickDisabled is true.

src/components/HashButton.tsx
```diff
@@ -57,9 +57,10 @@ function HashLink(hash: string, type: HashType): JSX.Element {
 interface HashButtonProps {
   hash: string;
   type: HashType;
+  alignRight?: boolean;
 }
 
-export default function HashButton({hash, type}: HashButtonProps) {
+export default function HashButton({hash, type, alignRight}: HashButtonProps) {
```
zacharydenton, line 63: I think it would be better to use spread props here (that way it can handle more than just aligning right).

src/components/Nav.tsx
```diff
@@ -16,6 +16,8 @@ function NavButton({
   title: string;
   label: string;
 }) {
+  const isSelected = window.location.pathname === to;
```
zacharydenton, line 19: I think this is built-in funtionality for NavLinks (https://reactrouter.com/en/main/components/nav-link).

src/pages/Validators/Table.tsx
```diff
@@ -140,6 +147,11 @@ export function ValidatorsTable({
   validators,
   columns = DEFAULT_COLUMNS,
 }: ValidatorsTableProps) {
+  const validatorsInOrder = validators.sort(
```
zacharydenton, line 150: It's not a good idea for a component to mutate its props. You should create a copy of the validators array and sort the copy.

src/components/GeneralTableRow.tsx
zihan-aptos: You're right, `onClick` would still work. Thanks for catching it. I love all your comments!!

msavir (approved):

PR #203, opened 2022-10-07
[Explorer] account transactions pagination

zihan-aptos:
https://user-images.githubusercontent.com/109111707/194495422-3d808080-2dc7-4e85-bf9f-59559332641b.mov

### DIFF
--- a/src/pages/Account/Tabs/TransactionsTab.tsx
+++ b/src/pages/Account/Tabs/TransactionsTab.tsx
@@ -1,27 +1,127 @@
-import React from "react";
+import React, {useState} from "react";
+import {Pagination, Box} from "@mui/material";
 import TransactionsTable from "../../Transactions/TransactionsTable";
 import Error from "../Error";
 import {useGetAccountTransactions} from "../../../api/hooks/useGetAccountTransactions";
 import EmptyTabContent from "../../../components/IndividualPageContent/EmptyTabContent";
+import {useGlobalState} from "../../../GlobalState";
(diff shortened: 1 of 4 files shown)


src/pages/Account/Tabs/TransactionsTab.tsx
```diff
@@ -1,33 +1,134 @@
...
+  sequenceNum: number;
 };
 
-export default function TransactionsTab({address}: TransactionsTabProps) {
-  const {isLoading, data, error} = useGetAccountTransactions(address);
+function TransactionsPaginationTable({
+  address,
+  sequenceNum,
+}: TransactionsPaginationTableProps) {
+  const [currentPageNum, setCurrentPageNum] = useState<number>(1);
```
zacharydenton, line 47: I recommend reusing the pagination logic from the other transactions table, including having `start` in the query string (instead of storing the page number as a state variable).

zihan-aptos: I thought about make `start` as a url param. However, this table is under a tab in the account page. Ideally we should have something like `/account/0x12345/transactions?start=123`. But since don't have the tab part `/transactions` yet, it'd be kinda weird to have `/account/0x12345?start=123` because other tabs in this page won't share the same `start`. 

Let me add a TODO here to add url param for both tabs and pagination. This PR is needed asap because without it users won't be able to see all their transactions. Thanks!

msavir (approved):

PR #208, opened 2022-10-10
[Explorer] fix warnings

zihan-aptos:
This PR fixed two nits. One was caused by including an icon component (a `<div>`) in a Typography (a `<p>`). The other was caused by querying block with transactions while transactions were not needed.

### DIFF
--- a/src/api/hooks/useGetBlock.ts
+++ b/src/api/hooks/useGetBlock.ts
@@ -4,23 +4,23 @@ import {getBlockByHeight, getBlockByVersion} from "../../api";
 import {ResponseError} from "../../api/client";
 import {useGlobalState} from "../../GlobalState";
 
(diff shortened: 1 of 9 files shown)


src/api/hooks/useGetBlock.ts
```diff
@@ -4,23 +4,23 @@ import {getBlockByHeight, getBlockByVersion} from "../../api";
 import {ResponseError} from "../../api/client";
 import {useGlobalState} from "../../GlobalState";
 
-export function useGetBlockByHeight(height: number) {
+export function useGetBlockByHeight(height: number, withTransactions = true) {
```
zacharydenton, line 7: I recommend using a keyword argument instead of a positional argument for this. That will make it clearer at the callsite what true/false means.

zacharydenton (approved):

PR #213, opened 2022-10-12
[Explorer] add network chain id

zihan-aptos:
* add chain ids in network select button
* hide local network when local is not available
* reorder networks
* remove "NETWORK" in home page header
Screen Shot 2022-10-12 at 3 27 41 PM

### DIFF
--- a/src/api/hooks/useGetNetworkChainIds.ts
+++ b/src/api/hooks/useGetNetworkChainIds.ts
@@ -1,0 +1,64 @@
+import {NetworkName, networks} from "../../constants";
+import {useQuery} from "react-query";
+import {getLedgerInfo} from "..";
+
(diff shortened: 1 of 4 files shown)


src/api/hooks/useGetNetworkChainIds.ts
```diff
@@ -0,0 +1,64 @@
...
+      );
+      const chainId: string | null = data?.chain_id
+        ? data?.chain_id.toString()
+        : null;
+
+      if (networkName !== "local" && chainId !== null) {
+        setWithExpiry(
+          `${networkName}ChainId`,
+          chainId,
+          3600000, // ttl is 1 hour
```
msavir, line 55: can be a const?

msavir (approved):

PR #227, opened 2022-10-18
[Explorer] Account not found

zihan-aptos:
Show error message on the page level (not tab level as previously) when account is not found.
Screen Shot 2022-10-17 at 5 45 17 PM

### DIFF
--- a/src/pages/Account/Error.tsx
+++ b/src/pages/Account/Error.tsx
@@ -1,7 +1,6 @@
 import React from "react";
 import {ResponseError, ResponseErrorType} from "../../api/client";
 import {Alert} from "@mui/material";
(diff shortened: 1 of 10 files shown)


src/pages/Account/Error.tsx
```diff
@@ -13,15 +12,15 @@ export default function Error({error, address}: ErrorProps) {
     return (
       <Alert severity="error">
         {error.message}
-        Could not find an Account with address {address}
+        Account not found with address {address}.
```
geekflyer, line 15:
```suggestion
        Account not found: {address}
```

geekflyer (approved):

PR #228, opened 2022-10-19
[Explorer] fix search bar

zihan-aptos:
### The problem:
When user type in something in the search bar, for example "0x1", the search bar fetch data for "0", "0x", and "0x1". It's possible that the results of "0" get returned after the results of "0x1" get returned. And the search bar will show the latest results that got returned which is the results of "0". This is not what we want since the user meant to search for "0x1".

### The solution:
There could be potentially a couple of solutions. I decided to do the following based on the performance and complexity of the implementation: we wait for the users to stop typing before fetching the search results. This way "0" and "0x1" won't be searched at all therefore less network requests and better performance.


msavir (approved):
Yes! debounce for server autocomplete search requests makes sense

PR #233, opened 2022-10-20
[Explorer] update address format for indexer queries

zihan-aptos:
When talking to the indexer, we should pass addresses as the format of `0x000...000123`. Before is `0x123` or `123`.

### DIFF
--- a/src/pages/Account/Tabs/TokensTab.tsx
+++ b/src/pages/Account/Tabs/TokensTab.tsx
@@ -25,9 +25,13 @@ type TokenTabsProps = {
 };
 
 export default function TokenTabs({address}: TokenTabsProps) {
+  // whenever talking to the indexer, the address needs to fill in leading 0s
+  // for example: 0x123 => 0x000...000123  (61 0s before 123)
+  const addr64Hash = "0x" + address.substring(2).padStart(64, "0");
+
   const {loading, error, data} = useQuery(TOKENS_QUERY, {
     variables: {
-      owner_address: address,
+      owner_address: addr64Hash,
     },
   });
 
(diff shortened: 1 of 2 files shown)


src/pages/Account/Tabs/TokensTab.tsx
```diff
@@ -25,9 +25,13 @@ type TokenTabsProps = {
 };
 
 export default function TokenTabs({address}: TokenTabsProps) {
+  // whenever talking to the indexer, the address needs to fill in leading 0s
+  // for example: 0x123 => 0x000...000123  (61 0s before 123)
+  const addr64Hash = "0x" + address.substring(2).padStart(64, "0");
```
msavir, line 30: I think we should be using the aptos sdk for that and encourage dApps to use the sdk for those kind of things

zihan-aptos: I tried `TxnBuilderTypes.AccountAddress.fromHex()` and the result is not a string. Do you know the best practice to convert it to a string?

msavir: cat try
```
const addressFromHex = TxnBuilderTypes.AccountAddress.fromHex(address);
const addressStr = HexString.fromUint8Array(BCS.bcsToBytes(addressFromHex)).hex();
```

zacharydenton: I agree in principle that we should rely on the SDK for this but right now it doesn't look like using the SDK is the better approach. Maybe in the future we can add a method to the SDK specifically for expanding an address to the full 32 bytes.

zihan-aptos: I'll add a TODO to track this. Thanks for the inputs.

bowenyang007 (approved):

src/pages/Account/Tabs/TokensTab.tsx
gregnazario: Yeah, let's add this to the SDK in the future

gregnazario (approved):

PR #239, opened 2022-10-25
Txn table

zihan-aptos:
Changes in this PR:
* "Status" column: remove this column and show a red error icon on the side of version id when the status is Fail.
* "Type" column: show icon only, added a tooltip modal to explain the meaning of these icons/types.
* "Hash" column: remove it, not informative in the transaction table
* "Timestamp" column: change the timestamp format to make it shorter
* "Sender" column: add this column to show the sender/proposer of a transaction
* "Send To" column: add this column to show either the receiver of a user transfer or the smart contract address this transaction interacts with, depending on the transaction type
* "Function" column: add this column to show the function for this transaction (what exactly is this transaction about)
* "Amount" column: add this column to show amount of a transaction, also merged Gas column into this one
  * for general transactions table, the amount shown is the amount that get involved in this transaction. the amount is neutral (no + or -).
  * for account transactions table, the amount shown is the specific amount that this account deposit/withdraw in this transaction. the amount can be + or -.
  
 Next Step:
* add tooltips for all the columns
Screen Shot 2022-10-26 at 11 10 15 AM
Screen Shot 2022-10-26 at 11 03 43 AM


src/components/TransactionType.tsx
```diff
@@ -43,6 +43,23 @@ function getTypeIcon(type: string, color?: Color) {
   }
 }
 
+function getTypeTooltip(type: string): string {
+  switch (type) {
+    case "block_metadata_transaction":
```
msavir, line 48: looks like its being used in different functions, wonder if it makes sense to extract it into an enum/type?

src/pages/Transactions/TransactionsTable.tsx
```diff
@@ -1,123 +1,213 @@
...
-    <TableCell sx={{textAlign: "left"}}>
-      {"success" in transaction && (
-        <TableTransactionStatus success={transaction.success} />
-      )}
-    </TableCell>
+    <GeneralTableCell sx={{textAlign: "left"}}>
+      <Stack direction="row" spacing={0.5}>
+        <Link
+          component={RRD.Link}
+          to={`/txn/${"version" in transaction && transaction.version}`}
```
msavir, line 54: what happens if there is no version?

msavir (approved):

src/components/TransactionType.tsx
zihan-aptos: yeah good call. let me add a TODO for it.

src/pages/Transactions/TransactionsTable.tsx
zihan-aptos: theoretically it's not possible. every transaction other than pending transaction has a version. but our blockchain is fast so we almost never see a pending transaction (i've never seen it, i doubt that it exists)

PR #249, opened 2022-10-28
[Explorer] show account ANS

zihan-aptos:
Screen Shot 2022-10-28 at 4 16 42 PM

### DIFF
--- a/src/api/hooks/useGetANS.ts
+++ b/src/api/hooks/useGetANS.ts
@@ -3,19 +3,16 @@ import {NetworkName} from "../../constants";
 import {useGlobalState} from "../../GlobalState";
 
 function getFetchNameUrl(network: NetworkName, address: string) {
-  switch (network) {
-    case "mainnet":
-      return `https://www.aptosnames.com/api/mainnet/v1/name/${address}`;
-    case "testnet":
-      return `https://www.aptosnames.com/api/testnet/v1/name/${address}`;
-    default:
-      return undefined;
+  if (network !== "testnet" && network !== "mainnet") {
+    return undefined;
   }
+
+  return `https://www.aptosnames.com/api/${network}/v1/name/${address}`;
 }
 
 export function useGetNameFromAddress(address: string) {
   const [state, _] = useGlobalState();
-  const [name, setName] = useState<string>();
+  const [name, setName] = useState<string | undefined>();
   const url = getFetchNameUrl(state.network_name, address);
 
   useEffect(() => {
@@ -26,7 +23,9 @@ export function useGetNameFromAddress(address: string) {
         setName(name);
       };
 
-      fetchData();
+      fetchData().catch((error) => {
+        console.error("ERROR!", error, typeof error);
+      });
     }
   }, [address, state]);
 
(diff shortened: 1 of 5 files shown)


src/api/hooks/useGetANS.ts
```diff
@@ -0,0 +1,65 @@
+import {useEffect, useState} from "react";
+import {NetworkName} from "../../constants";
+import {useGlobalState} from "../../GlobalState";
+
+function getFetchNameUrl(network: NetworkName, address: string) {
```
msavir, line 5: nite:
```suggestion
function getFetchNameUrl(network: NetworkName, address: string) { 
if(network !== "testnet" && network !== "mainnet") return undefined;
return `https://www.aptosnames.com/api/${network}/v1/name/${address}`
}
```

src/components/TitleHashButton.tsx
```diff
@@ -80,3 +98,31 @@ export default function HashButtonCopyable({hash}: HashButtonCopyableProps) {
...
+          height: BUTTON_HEIGHT,
+          backgroundColor: `${
+            theme.palette.mode === "dark" ? grey[600] : grey[200]
+          }`,
+          borderRadius: 1,
+          color: "inherit",
+          padding: "0.15rem 1rem 0.15rem 1rem",
+        }}
+      >
+        <Typography variant="body2">{name}</Typography>
```
msavir: should we show `.apt`? `{label}.apt`

src/api/hooks/useGetANS.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export function useGetAddressFromName(name: string) {
+  const [state, _] = useGlobalState();
+  const [address, setAddress] = useState<string>();
+  const url = getFetchAddressUrl(state.network_name, name);
+
+  useEffect(() => {
+    if (url !== undefined) {
+      const fetchData = async () => {
+        const response = await fetch(url);
```
msavir, line 55: maybe worth using `try{}catch(){}` for server errors

src/components/TitleHashButton.tsx
```diff
@@ -80,3 +106,37 @@ export default function HashButtonCopyable({hash}: HashButtonCopyableProps) {
...
+          color: "inherit",
+          padding: "0.15rem 1rem 0.15rem 1rem",
+        }}
+      >
+        <Link
+          href={`https://www.aptosnames.com/name/${name}`}
+          target="_blank"
+          underline="none"
+        >
+          {name}
```
msavir, line 137: should we show `.apt`? `${name}.apt`?

src/api/hooks/useGetANS.ts
```diff
@@ -0,0 +1,65 @@
...
+    case "testnet":
+      return `https://www.aptosnames.com/api/testnet/v1/name/${address}`;
+    default:
+      return undefined;
+  }
+}
+
+export function useGetNameFromAddress(address: string) {
+  const [state, _] = useGlobalState();
+  const [name, setName] = useState<string>();
```
msavir, line 18: I guess that can be `string | null/undefined`...?

src/api/hooks/useGetANS.ts
```diff
@@ -0,0 +1,65 @@
...
+    case "mainnet":
+      return `https://www.aptosnames.com/api/mainnet/v1/address/${name}`;
+    case "testnet":
+      return `https://www.aptosnames.com/api/testnet/v1/address/${name}`;
+    default:
+      return undefined;
+  }
+}
+
+export function useGetAddressFromName(name: string) {
```
msavir, line 47: is it being used anywhere...?

src/components/HashButton.tsx
```diff
@@ -83,6 +132,9 @@ export default function HashButton({
   const open = Boolean(anchorEl);
   const id = open ? "simple-popover" : undefined;
```
msavir, line 133: maybe meaningful popover name?

src/components/TitleHashButton.tsx
zihan-aptos: I would argue that it's not needed? For example etherscan does not show ".eth". Also I'm following Petra who shows the name only.

src/api/hooks/useGetANS.ts
zihan-aptos: It'll only be `string | undefined`, and with the default value unset the type is inferred as `string | undefined`. but let me add it.

src/api/hooks/useGetANS.ts
zihan-aptos: Right now not yet. I'll work on another PR that enables searching names in the search bar. It'll be used soon.

src/components/HashButton.tsx
zihan-aptos: sorry what do you mean? also this is copied from Joe's code I actually don't know what it is lol

msavir: I think its the id for the popover, just saw it here so thought a meaningful name might be better

msavir (approved):

PR #251, opened 2022-10-31
[Explorer] search aptos names

zihan-aptos:
Screen Shot 2022-10-31 at 11 43 27 AM

### DIFF
--- a/src/api/hooks/useGetANS.ts
+++ b/src/api/hooks/useGetANS.ts
@@ -59,3 +59,26 @@ export function useGetAddressFromName(name: string) {
 
   return address;
 }
+
+export async function getAddressFromName(
+  name: string,
+  network: NetworkName,
+): Promise<string | undefined> {
+  const addressUrl = getFetchAddressUrl(network, name);
+  if (addressUrl === undefined) {
+    return undefined;
+  }
+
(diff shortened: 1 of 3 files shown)


msavir (approved):

src/api/hooks/useGetANS.ts
```diff
@@ -59,3 +59,26 @@ export function useGetAddressFromName(name: string) {
...
+export async function getAddressFromName(
+  name: string,
+  network: NetworkName,
+): Promise<string | undefined> {
+  const addressUrl = getFetchAddressUrl(network, name);
+  if (addressUrl === undefined) {
+    return undefined;
+  }
+
+  const addressResponse = await fetch(addressUrl);
```
msavir, line 72: should we add `catch`?

zihan-aptos: Hmm... I can't think of a case when it'll throw error, unless the url has changed.

zihan-aptos: Oh and the upper level where this function is called has try catch. So it should be safe.

msavir: if the server is down for some reason...?

PR #257, opened 2022-11-01
[Explorer] Add peak TPS

zihan-aptos:
Screen Shot 2022-11-01 at 12 17 53 PM
Screen Shot 2022-11-01 at 12 16 11 PM

### DIFF
--- a/src/pages/LandingPage/NetworkInfo/MetricCard.tsx
+++ b/src/pages/LandingPage/NetworkInfo/MetricCard.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import {Typography, Stack} from "@mui/material";
+import {Typography, Stack, useTheme} from "@mui/material";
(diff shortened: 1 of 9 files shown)


src/pages/LandingPage/NetworkInfo/MetricCard.tsx
```diff
@@ -1,39 +1,95 @@
...
+export default function MetricCard({data, label, tooltip}: MetricCardProps) {
+  return (
+    <MetricCardComponent label={label} tooltip={tooltip}>
+      <Data>{data}</Data>
+    </MetricCardComponent>
+  );
+}
+
+type DoubleMetricCardProps = {
+  data1: string;
```
msavir, line 65: what are those data/label represent? can we think on better naming?

msavir (approved):

PR #274, opened 2022-11-18
[Validators Page] set expiration time for localstorage

0xZihan:


src/api/hooks/useGetValidatorSet.ts
```diff
@@ -95,7 +101,9 @@ export function useGetValidatorToOperator() {
   // to avoid hitting rate limiting
   // as the query is very expensive
   useEffect(() => {
-    const validatorToOperatorStr = localStorage.getItem("validatorToOperator");
+    const validatorToOperatorStr = getLocalStorageWithExpiry(
+      "validatorToOperatorWithExpiry",
```
msavir, line 105: can use it as a const instead

msavir (approved):

PR #276, opened 2022-11-30
[Analytics] set up Analytics page and add line chart

0xZihan:
* set up Analytics page
* add Daily Active User line chart
* remove unused files (Card and AccountInfo etc.)
Screen Shot 2022-11-30 at 3 54 43 PM


src/pages/Analytics/Components/ChartRangeDaysSelect.tsx
```diff
@@ -0,0 +1,35 @@
...
+  days,
+  setDays,
+}: ChartRangeDaysSelectProps) {
+  const handleChange = (event: SelectChangeEvent) => {
+    setDays(parseInt(event.target.value));
+  };
+
+  return (
+    <Select
+      id="chart-range-days-select"
```
msavir, line 25: is this `id` needed?

src/api/hooks/useGetAnalyticsData.ts
```diff
@@ -0,0 +1,68 @@
...
+  const [data, setData] = useState<AnalyticsData>();
+
+  useEffect(() => {
+    const fetchData = async () => {
+      const response = await fetch(DATA_URL);
+      const data = await response.json();
+      setData(data);
+    };
+
+    if (state.network_name === defaultNetworkName) {
```
msavir, line 58: is this `if` is because we want it to error only on mainnet? can add comments around this check?

msavir (approved):

PR #290, opened 2022-12-13
Disable "Maintain Aspect Ratio" for Charts

moonclavedev:
Disables `maintainAspectRatio` for Charts, providing more consistent results upon browser resize.


0xZihan (approved):
Thanks for jumping in!

PR #297, opened 2023-01-05
[explorer] add skeleton loading on validator page 

jillxuu:
Fixes WP-282: we had null or undefined value while data's loading in UI rendering. this pr uses react loading skeleton to make the UI prettier and user friendly. 

 **was debating which approach to go with:**
1. create a skeleton component at the top level so that if any of the content data not ready we directly load skeleton compoenent
2. create some skeleton constants for map metrics sections since these datas take some time to render
3. directly use skeleton in each row

end up deciding to go with approach 2 cuz approach 1 has too much duplicate of work and if we ever wanna make some change in the future, we might need to change in multiple places. approach 3 doesn't have a single place for skeleton effect and might be difficult to find all of them efficiently for the future work. 

https://user-images.githubusercontent.com/121921928/210890256-e44ae561-ea37-4349-8eef-5edf66188b43.mov

https://user-images.githubusercontent.com/121921928/210890228-46c2d144-edfc-4933-b14c-7b614a4a2141.mov

https://user-images.githubusercontent.com/121921928/210890250-a9dee0de-351f-4450-87b7-e0bcf7aa819c.mov

https://user-images.githubusercontent.com/121921928/210890240-188e839b-f465-467e-b7c1-909f56919f58.mov


src/pages/Validators/ValidatorsMap.tsx
```diff
@@ -1,38 +1,71 @@
...
+      totalVotingPower &&
+      rewardsRateYearly &&
+      numberOfActiveValidators
+    ) {
+      setIsSkeletonLoading(false);
+    }
+  }, [curEpoch, totalVotingPower, rewardsRateYearly, numberOfActiveValidators]);
+
+  return (
+    <SkeletonTheme baseColor={isDarkTheme ? grey[500] : undefined}>
```
0xZihan, line 37: what's `baseColor` here?

src/pages/Validators/ValidatorsMap.tsx
```diff
@@ -1,38 +1,71 @@
...
+          direction="column"
+          justifyContent="space-between"
+          marginY={4}
+          sx={{backgroundColor: backgroundColor}}
+          overflow="hidden"
+        >
+          <Map validatorGeoGroups={validatorGeoGroups} />
+          <MapMetrics
+            validatorGeoMetric={validatorGeoMetric}
+            isOnMobile={isOnMobile}
```
0xZihan, line 49:
```suggestion
            isOnMobile
```

src/pages/Validators/ValidatorsMap.tsx
```diff
@@ -1,38 +1,71 @@
...
+        <Stack
+          direction="row"
+          justifyContent="space-between"
+          marginY={4}
+          sx={{backgroundColor: backgroundColor}}
+          overflow="hidden"
+        >
+          <MapMetrics
+            validatorGeoMetric={validatorGeoMetric}
+            isOnMobile={isOnMobile}
```
0xZihan, line 63:
```suggestion
            isOnMobile
```

src/pages/Validators/ValidatorsMap.tsx
```diff
@@ -1,38 +1,71 @@
...
+            isSkeletonLoading={isSkeletonLoading}
```
0xZihan, line 64:
```suggestion
            isSkeletonLoading
```

src/pages/Validators/ValidatorsMap.tsx
```diff
@@ -1,38 +1,71 @@
...
+          justifyContent="space-between"
+          marginY={4}
+          sx={{backgroundColor: backgroundColor}}
+          overflow="hidden"
+        >
+          <Map validatorGeoGroups={validatorGeoGroups} />
+          <MapMetrics
+            validatorGeoMetric={validatorGeoMetric}
+            isOnMobile={isOnMobile}
+            isSkeletonLoading={isSkeletonLoading}
```
0xZihan, line 50:
```suggestion
            isSkeletonLoading
```

src/pages/Validators/constants.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <Skeleton width={150} />
+    </Typography>
+    <Typography sx={{fontSize: {xs: fontSizeBodySmall, md: fontSizeSubtitle}}}>
+      <Skeleton width={140} />
+    </Typography>
+    <Typography sx={{fontSize: {xs: fontSizeBodySmall, md: fontSizeSubtitle}}}>
+      <Skeleton width={130} />
+    </Typography>
+  </MetricSection>
+);
```
0xZihan, line 42: react components should not be living in `constants`. Let's move them out and keep `constants.ts`.

src/pages/Validators/constants.tsx
```diff
@@ -0,0 +1,42 @@
...
+import React from "react";
+import MetricSection from "./Components/MetricSection";
+
+export const fontSizeBody = 14;
+export const fontSizeBodySmall = 12;
+export const fontSizeSubtitle = 17;
+export const fontSizeSubtitleSmall = 15;
+export const fontSizeTitle = 25;
+export const fontSizeTitleSmall = 15;
+export const epochSectionSkeleton = (
```
0xZihan, line 11: if you're trying to avoid having too many files, this component `EpochSectionSkeleton` can live in `EpochSection.tsx`.  This way `EpochSection()` can simply take `isLoading` as a prop, and render skeleton or the actual component based on `isLoading`. This can also encapsulate everything about EpochSection within one file/component.

src/pages/Validators/constants.tsx
```diff
@@ -0,0 +1,42 @@
...
+  <MetricSection>
+    <Stack direction="row" spacing={1} alignItems="center">
+      <Skeleton width={140} />
+    </Stack>
+    <Skeleton width={180} />
+    <Skeleton width={180} />
+  </MetricSection>
+);
+export const stakingSectionSkeleton = (
```
0xZihan, line 20: similar treatment for `StakingSection`

src/pages/Validators/constants.tsx
```diff
@@ -0,0 +1,42 @@
...
+  <MetricSection>
+    <Stack direction="row" spacing={1} alignItems="center">
+      <Skeleton width={190} />
+    </Stack>
+    <Stack direction="row" spacing={1} alignItems="center">
+      <Skeleton width={150} />
+    </Stack>
+  </MetricSection>
+);
+export const nodeCountsSectionSkeleton = (
```
0xZihan, line 30: similar treatment. you can take `nodeCountSection` out to make its own file/component like `EpochSection` and `StakingSection`.

0xZihan (changes requested):
Very impressive work for you first PR!! Only a few code style suggestions then we're good to go.

src/pages/Validators/ValidatorsMap.tsx
jillxuu: the reason why i specify the prop value here even though they have same naming, is cuz if i don't do so, the child component doesn't seem to be picking up parent's prop value change, see this attachment : Screenshot 2023-01-05 at 4 45 49 PM. the prop we pass down to MapMetrics was not updated accordingly.

src/pages/Validators/ValidatorsMap.tsx
jillxuu: ditto

src/pages/Validators/ValidatorsMap.tsx
jillxuu: ditto

src/pages/Validators/ValidatorsMap.tsx
0xZihan: Interesting. This shouldn't happen. It worked well before. see https://github.com/aptos-labs/explorer/blob/main/src/pages/Validators/ValidatorsMap.tsx#L24. Can you dig deeper and try to see why? It's not a blocker though.

src/pages/Validators/ValidatorsMap.tsx
jillxuu: this's the skeleton theme's loading color, which should match the theme we have either light or dark.

src/pages/Validators/ValidatorsMap.tsx
zacharydenton: FYI, `<Component isOnMobile>` is shorthand for `<Component isOnMobile={true}>`, not `<Component isOnMobile={isOnMobile}>`. Think HTML boolean attributes, not ES property names.

PR #302, opened 2023-01-11
[explorer][delegation] add the validator page

jillxuu:
* add the validator page that's for a single validator. still in early development and have many more improvements to make. but would like to send pr out to iterate while we're still finalizing on designs, validator info and staking flows.
* uses `inDev` to gate the change. 

https://user-images.githubusercontent.com/121921928/211883346-8fd6ec74-8f09-4a89-9ba9-d5256045a3ae.mov


package.json
```diff
@@ -25,6 +25,7 @@
     "moment": "^2.29.1",
     "prettier": "^2.6.2",
     "react": "latest",
+    "react-adobe-fonts": "^0.1.0",
```
0xZihan, line 28: why do we need this?

src/components/IndividualPageContent/ContentRow.tsx
```diff
@@ -9,22 +9,27 @@ type ContentRowProps = {
   value: React.ReactNode;
   tooltip?: React.ReactNode;
   i?: any;
+  container?: boolean;
+  sx?: SxProps<Theme>;
 };
 
 export default function ContentRow({
```
0xZihan, line 16: do you mind holding off the style change in this component? I think it's a design discussion that do we want to introduce different content row style for the validator page. If there's any change, let's do it after the design gets finalized.

src/pages/Validators/ValidatorsTable.tsx
```diff
@@ -256,15 +259,17 @@ type ValidatorRowProps = {
   validator: MainnetValidator;
   validatorToOperator: {[name: string]: string} | null;
   columns: Column[];
+  onClick?: () => void;
```
0xZihan, line 262: we don't need to pass in onClick here

src/pages/Validators/ValidatorsTable.tsx
```diff
@@ -256,15 +259,17 @@ type ValidatorRowProps = {
...
 
 function ValidatorRow({
   validator,
   validatorToOperator,
   columns,
+  onClick,
 }: ValidatorRowProps) {
   return (
-    <GeneralTableRow>
+    <GeneralTableRow onClick={onClick}>
```
0xZihan, line 272: just do `onClick={inDev ? () => rowClick(validator.address) : () => null}`

src/pages/Validators/ValidatorsTable.tsx
```diff
@@ -320,6 +331,7 @@ export function ValidatorsTable({
               validator={validator}
               validatorToOperator={validatorToOperator}
               columns={columns}
+              onClick={inDev ? () => rowClick(validator.address) : () => null}
```
0xZihan, line 334: so that we can remove it here

src/pages/Validator/DetailCard.tsx
```diff
@@ -0,0 +1,114 @@
...
+} from "../../api/hooks/useGetValidatorSet";
+import {useGetAccountResource} from "../../api/hooks/useGetAccountResource";
+import {ValidatorCardDetailName, validatorCardDetailType} from "./Constants";
+import {useEffect, useState} from "react";
+
+type ValidatorDetailProps = {
+  address: Types.Address;
+};
+
+export default function ValidatorDetailCard({address}: ValidatorDetailProps) {
```
0xZihan, line 23: It's very nice to explore different coding styles! Good job! However, for consistency across the code base, I would recommend implementing the content card/tab in similar style. Take a look at `UserTransactionOverviewTab` and places that uses `<ContentRow .../>`.

src/pages/Validator/DetailCard.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const sx = {display: "flex", justifyContent: "space-between"};
+  const [isSkeletonLoading, setIsSkeletonLoading] = useState<boolean>(true);
+
+  const lockedUntilSecs = accountResource
+    ? BigInt((accountResource.data as any).locked_until_secs)
+    : null;
+  const operatorAddr = validatorToOperator
+    ? validatorToOperator[addressHex]
+    : null;
+  const rewardGrowth = validator ? validator.rewards_growth : null;
```
0xmaayan, line 44: I think you simplify it and do

```suggestion
  const rewardGrowth = validator?.rewards_growth
```

src/pages/utils.tsx
```diff
@@ -112,3 +112,14 @@ export function getTableFormattedTimestamp(timestamp?: string): string {
 
   return timestamp_display.local_formatted_short;
 }
+
+export function prettifyTimestamp(timestamp: number) {
```
0xmaayan, line 116: can add comments on what the format we expect to get when calling this function

yarn.lock
```diff
@@ -7874,6 +7874,11 @@ raw-body@2.5.1:
     iconv-lite "0.4.24"
     unpipe "1.0.0"
 
+react-adobe-fonts@^0.1.0:
```
0xmaayan, line 7877: is it needed?

package.json
jillxuu: you're right we don't need it. somehow it got added in here.

src/components/IndividualPageContent/ContentRow.tsx
jillxuu: yes in the new revision, will remove these design changes!

src/pages/Validators/ValidatorsTable.tsx
jillxuu: yes you're right!

src/pages/Validator/DetailCard.tsx
jillxuu: ofc! initially was thinking how to write shorter code with concise render function but consistency def matters. will change it back!

package.json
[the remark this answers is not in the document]
jillxuu: nah it's not used anywhere. somehow it got added here. will remove it!

src/pages/Validator/DetailCard.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const {validators} = useGetMainnetValidators();
+  const {isLoading, accountResource} = useGetAccountResource(
+    addressHex,
+    "0x1::stake::StakePool",
+  );
+  const validatorToOperator = useGetValidatorToOperator();
+  const validator = validators.find(
+    (validator) => validator.address === addressHex,
+  );
+  const sx = {display: "flex", justifyContent: "space-between"};
```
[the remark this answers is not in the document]
jillxuu: cuz initially i was using `contentRow` for each entry and sx with same props was used multiple places.

yarn.lock
jillxuu: nah! will delete it

PR #307, opened 2023-01-17
[explorer][delegation] add the staking bar on the validator page

jillxuu:
* create this bar in bare minimum. will add further UI and data source later. its gated in the dev mode
Screenshot 2023-01-17 at 11 31 05 AM

### DIFF
--- a/src/pages/Validator/index.tsx
+++ b/src/pages/Validator/index.tsx
@@ -6,19 +6,18 @@ import ValidatorTitle from "./Title";
 import ValidatorDetailCard from "./DetailCard";
 import ValidatorStakingBar from "./StakingBar";
 import {useGetMainnetValidators} from "../../api/hooks/useGetValidatorSet";
-import EmptyValue from "../../components/IndividualPageContent/ContentValue/EmptyValue";
+import {HexString} from "aptos";
 
 export default function ValidatorPage() {
   const address = useParams().address ?? "";
-  // make sure that addresses will always start with "0x"
-  const addressHex = address.startsWith("0x") ? address : "0x" + address;
+  const addressHex = new HexString(address);
   const {validators} = useGetMainnetValidators();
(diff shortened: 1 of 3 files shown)


src/pages/Validator/index.tsx
```diff
@@ -4,17 +4,31 @@ import {useParams} from "react-router-dom";
...
 import ValidatorTitle from "./Title";
 import ValidatorDetailCard from "./DetailCard";
+import ValidatorStakingBar from "./StakingBar";
+import {useGetMainnetValidators} from "../../api/hooks/useGetValidatorSet";
+import EmptyValue from "../../components/IndividualPageContent/ContentValue/EmptyValue";
 
 export default function ValidatorPage() {
   const address = useParams().address ?? "";
+  // make sure that addresses will always start with "0x"
+  const addressHex = address.startsWith("0x") ? address : "0x" + address;
```
0xmaayan, line 14: should this be a global util func? I think I've seen it somewhere else also. another option is to utilize the ts sdk and do `const addressHex = new HexString(address);`
https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/hex_string.ts#L65

src/pages/Validator/index.tsx
```diff
@@ -4,17 +4,31 @@ import {useParams} from "react-router-dom";
...
+  const {validators} = useGetMainnetValidators();
```
0xmaayan, line 15: can `validators` be null/undefined on first load? do we need to use `useEffect` here?

0xmaayan (approved):

src/pages/Validator/index.tsx
jillxuu: good point! but instead of using useEffect, we can use skeleton to render the loading page. will add it in the following pr!

PR #308, opened 2023-01-18
[Validators Page] data BE

0xZihan:
tl;dr: Right now data in [the validators page](https://explorer.aptoslabs.com/validators) comes from 3 sources. We want to unify them all together to one data source for better performance and accuracy.

## Problem:
As mentioned above, the 3 sources are:
* validators liveness data (old leaderboard): https://aptos-analytics-data-mainnet.s3.amazonaws.com/liveness.json
* validators geo related data: https://aptos-analytics-data-mainnet.s3.amazonaws.com/validator_location_stats.json
* validators staking_pool_address => operator_address mapping: data directly from the chain with some local storage performance optimization.

## Solution:
In this PR, we make all these data available in one query.

## No observable change for users


src/api/hooks/useGetMainnetValidators.ts
```diff
@@ -0,0 +1,76 @@
...
+  return {validatorsRawData};
+}
+
+export function useGetMainnetValidators() {
+  const {activeValidators} = useGetValidatorSet();
+  const {validatorsRawData} = useGetMainnetValidatorsRawData();
+  const [validators, setValidators] = useState<MainnetValidatorData[]>([]);
+
+  useEffect(() => {
+    if (activeValidators.length > 0 && validatorsRawData.length > 0) {
```
jillxuu, line 61: do we wanna check whether both `activeValidators` and `validatorsRawData` have equal length? we had sev about this previously cuz we added new validator to the list so that we couldn't find the match for that newly added validator?

src/pages/Validator/DetailCard.tsx
```diff
@@ -7,38 +7,32 @@ import {HashType} from "../../components/HashButton";
...
 import TimestampValue from "../../components/IndividualPageContent/ContentValue/TimestampValue";
+import {useGetMainnetValidators} from "../../api/hooks/useGetMainnetValidators";
 
 type ValidatorDetailProps = {
   address: Types.Address;
 };
 
 export default function ValidatorDetailCard({address}: ValidatorDetailProps) {
-  // make sure that addresses will always start with "0x"
   const addressHex = address.startsWith("0x") ? address : "0x" + address;
```
jillxuu, line 20: maayan left a [comment](https://github.com/aptos-labs/explorer/pull/307#discussion_r1072971727) on one of my PRs saying that we could use sdk for converting address to hex string. we can do the same thing here.

src/pages/Validator/DetailCard.tsx
```diff
@@ -7,38 +7,32 @@ import {HashType} from "../../components/HashButton";
...
   );
   const [isSkeletonLoading, setIsSkeletonLoading] = useState<boolean>(true);
-
   const lockedUntilSecs = accountResource
     ? BigInt((accountResource.data as any).locked_until_secs)
     : null;
-  const operatorAddr = validatorToOperator
-    ? validatorToOperator[addressHex]
-    : null;
+  const operatorAddr = validator?.operator_address;
```
jillxuu, line 35: good catch!

jillxuu (approved):
changes for combining couple data sources into one! def make the data consumption a lot more straightforward! before we land this diff, could we verify the explorer's working properly by unploading the video recording to the description section?

PR #310, opened 2023-01-18
[explorer][delegaiton] update validators page with the new delegatable tab

jillxuu:
create the delegation node tab on the validators page and make sure it's properly gated.
will improve UI in future pr. 

https://user-images.githubusercontent.com/121921928/213314966-9d5ad0c5-74e7-4248-9b2e-8799ee22b8cb.mov

### DIFF
--- a/src/ExplorerRoutes.tsx
+++ b/src/ExplorerRoutes.tsx
@@ -12,22 +12,17 @@ import BlocksPage from "./pages/Blocks/Index";
 import ValidatorsPage from "./pages/Validators/Index";
 import ValidatorPage from "./pages/Validator";
 import AnalyticsPage from "./pages/Analytics/Index";
-import {useGetInDevMode} from "./api/hooks/useGetInDevMode";
 
 export default function ExplorerRoutes() {
-  const inDev = useGetInDevMode();
-
   return (
     <ExplorerLayout>
       <Routes>
         <Route path="/" element={<LandingPage />} />
         <Route path="/transactions" element={<TransactionsPage />} />
         <Route path="/validators" element={<ValidatorsPage />} />
-        {inDev ? (
-          <Route path="/validator">
-            <Route path=":address" element={<ValidatorPage />} />
-          </Route>
-        ) : null}
+        <Route path="/validator">
+          <Route path=":address" element={<ValidatorPage />} />
+        </Route>
         <Route path="/txn">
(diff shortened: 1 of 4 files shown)


src/ExplorerRoutes.tsx
```diff
@@ -12,22 +12,17 @@ import BlocksPage from "./pages/Blocks/Index";
...
     <ExplorerLayout>
       <Routes>
         <Route path="/" element={<LandingPage />} />
         <Route path="/transactions" element={<TransactionsPage />} />
         <Route path="/validators" element={<ValidatorsPage />} />
-        {inDev ? (
-          <Route path="/validator">
-            <Route path=":address" element={<ValidatorPage />} />
-          </Route>
-        ) : null}
```
jillxuu, line 30: no need to have this `inDev` logic check here cuz we already gated one layer up in validators table row click.

src/pages/Validators/Index.tsx
```diff
@@ -8,6 +10,15 @@ import {ValidatorsTable} from "./ValidatorsTable";
 
 export default function ValidatorsPage() {
   const [state, _] = useGlobalState();
+  const [onDelegatory, setOnDelegatory] = useState<boolean>(false);
+  const inDev = useGetInDevMode();
+
+  const handleTabChange = (
+    _event: React.SyntheticEvent,
```
0xmaayan, line 17: is the `_event` needed?

src/pages/Validators/Index.tsx
```diff
@@ -18,14 +29,28 @@ export default function ValidatorsPage() {
       {state.network_name === "mainnet" ? (
         <>
           <ValidatorsMap />
+          {inDev ? (
```
0xmaayan, line 32:
```suggestion
          {inDev && ( 
           <Tabs value={onDelegatory} onChange={handleTabChange}>
              <Tab label="All Nodes" value={false} />
              <Tab label="Delegation Nodes" value={true} />
            </Tabs>
          )
```

0xmaayan (approved):

src/pages/Validators/Index.tsx
jillxuu: yes, event itself is not used in this `handleTabChange` function, but it's needed in `<Tabs onChange={xx}>` cuz `event` is: "The event source of the callback" to be consumed by `onChange`

PR #312, opened 2023-01-19
[explorer][delegation] refactor wallet connection flow

jillxuu:
this is just to make the flow work, since we copy pasted many logics directly from [ANS codebase](https://github.com/aptos-labs/aptos-names/tree/main/src/components/WalletConnector). a better way of handling this is to update wallet adaptor's UI package, which is already in my todo list. hopefully i can get it done by next week.

https://user-images.githubusercontent.com/121921928/213340827-05ab3149-45eb-4b81-abfa-ceb6d69ba88d.mov

### DIFF
--- a/src/pages/layout/ConnectWallet.tsx
+++ b/src/pages/layout/ConnectWallet.tsx
@@ -1,6 +1,0 @@
-import React from "react";
-import {WalletSelector} from "@aptos-labs/wallet-adapter-ant-design";
-
-export default function Wallet() {
-  return <WalletSelector />;
-}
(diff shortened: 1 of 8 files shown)


src/pages/layout/Header.tsx
```diff
@@ -121,7 +121,7 @@ export default function Header() {
               {theme.palette.mode === "light" ? <IconLight /> : <IconDark />}
             </Button>
             <NavMobile />
-            {inDev ? <ConnectWallet /> : null}
+            <Box sx={{marginLeft: "1rem"}}>{inDev && <WalletConnector />}</Box>
```
0xmaayan, line 124: Should we hide the Box component also if inDev mode?

src/pages/layout/ConnectWallet.tsx
```diff
@@ -1,6 +0,0 @@
-import React from "react";
-import {WalletSelector} from "@aptos-labs/wallet-adapter-ant-design";
```
0xmaayan, line 2: Can probably remove this package from package.json also

0xmaayan (approved):

PR #313, opened 2023-01-19
[explorer][delegation] refactor validators page UI

jillxuu:
refactor UI to match figma design

https://user-images.githubusercontent.com/121921928/213544888-6d894ea4-7de6-46b4-813c-d27174b788fd.mov

https://user-images.githubusercontent.com/121921928/213827660-d5d8ff41-9d10-4bfe-a497-c7c1826c4d95.mov


src/pages/Validators/Index.tsx
```diff
@@ -30,23 +30,37 @@ export default function ValidatorsPage() {
         <>
           <ValidatorsMap />
           {inDev && (
-            <Tabs value={onDelegatory} onChange={handleTabChange}>
-              <Tab label="All Nodes" value={false} />
-              <Tab label="Delegation Nodes" value={true} />
-            </Tabs>
+            <ToggleButtonGroup
+              sx={{width: 1 / 3, height: 50}}
```
zacharydenton, line 34: How does width: 1/3 look on mobile?

jillxuu: good call, on mobile, text gets to be two lines. what would be a mobile friendly way to set the size and width on these buttons?

zacharydenton: Personally I would use flexbox and a media query but it looks like the way to do it in emotion is https://theme-ui.com/sx-prop#responsive-values.

0xZihan:
It belongs to its own PR but can we hide the wallet button when it's on mobile (small screen)?

jillxuu:
@0xZihan yes, mobile friendly refactoring is on the todo list in future prs!

0xZihan (changes requested):

0xZihan:
> @0xZihan yes, mobile friendly refactoring is on the todo list in future prs!

awesome thanks!

src/pages/Validators/Table.tsx
```diff
@@ -80,6 +80,22 @@ function NetworkAddrCell({validator}: ValidatorCellProps) {
   );
 }
 
+function DelegatedStakeAmountCell({validator}: ValidatorCellProps) {
```
0xZihan, line 83: Wondering why do we need to change this file `Table.tsx` at all? From what I understand, the "old" validators table should not be changed at all for the delegation project.

src/pages/Validators/Index.tsx
```diff
@@ -30,23 +30,63 @@ export default function ValidatorsPage() {
...
         </>
       ) : (
         <>
-          {inDev ? (
-            <Tabs value={onDelegatory} onChange={handleTabChange}>
-              <Tab label="All Nodes" value={false} />
-              <Tab label="Delegation Nodes" value={true} />
-            </Tabs>
-          ) : null}
+          {inDev && (
```
0xZihan, line 64: Also to avoid code duplication, consider to change the logic structure. For example:

```
const validatorsTable = isMainnet ? <ValidatorsTable /> : <OldValidatorsTable />;

return inDev ? <ValidatorsTabs /> : validatorsTable;
```

Then inside `ValidatorsTabs`:
```
const validatorsTable = isMainnet ? <ValidatorsTable /> : <OldValidatorsTable />;

// the following is not real code. it simply means two tables under two tabs
return <Tabs>
  {validatorsTable}
  <DelegatoryValidatorsTable />
</Tabs>
```

Then inside `DelegatoryValidatorsTable`, this is the actual table component for delegatory validators. And this table will be shared between mainnet/testnet/devnet.

Of course this is only one of the ways that we can implement it. Please let me know if you have questions.

src/pages/Validators/Table.tsx
jillxuu: old table.tsx is for non-mainnet network. i had the assumption that since we are testing from devnet all the way to mainnet, we will need to update non-mainnet network validators table as well so that we know data looks about right! if that's not the case, i'll get this piece out of the pr! lmk

src/pages/Validators/Index.tsx
```diff
@@ -30,23 +30,63 @@ export default function ValidatorsPage() {
         <>
           <ValidatorsMap />
           {inDev && (
-            <Tabs value={onDelegatory} onChange={handleTabChange}>
-              <Tab label="All Nodes" value={false} />
-              <Tab label="Delegation Nodes" value={true} />
-            </Tabs>
+            <ToggleButtonGroup
```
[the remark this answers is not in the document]
jillxuu: makes tons of sense to me, thanks for flagging this!

src/pages/Validators/ValidatorsTable.tsx
```diff
@@ -158,10 +169,14 @@ function ValidatorHeaderCell({
       );
     case "delegator":
       return <GeneralTableHeaderCell header="Delegators" />;
+    case "rewardsEarned":
+      return <GeneralTableHeaderCell header="Rewards Earned" />;
     case "commission":
       return <GeneralTableHeaderCell header="Commission" />;
     case "myDeposit":
```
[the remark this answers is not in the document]
jillxuu: i see your point, these data are not part of "stage 1" in the milestones, but we will eventually need it next week when indexer's ready to have relevant data to serve us? is it okay to keep them in the table? or there's alternative understanding on "not in the mvp"? lmk!

src/pages/Validators/Table.tsx
```diff
@@ -162,9 +179,16 @@ function ValidatorHeaderCell({column}: ValidatorHeaderCellProps) {
     case "delegator":
       return <GeneralTableHeaderCell header="Delegators" textAlignRight />;
     case "commission":
-      return <GeneralTableHeaderCell header="Commission" textAlignRight />;
+      return <GeneralTableHeaderCell header="Commission" />;
     case "myDeposit":
```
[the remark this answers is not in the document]
jillxuu: same reasoning as [https://github.com/aptos-labs/explorer/pull/313#discussion_r1084853576](here). lmk!

src/pages/Validators/Table.tsx
0xZihan: For non-delegation nodes, we shouldn't change anything to the table. It's the case for all networks. So basically, if you think about it, we're adding a delegation nodes table for all 3 networks, but not changing anything to the existing ones.

src/pages/Validators/ValidatorsTable.tsx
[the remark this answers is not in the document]
0xZihan: I like clean code so I'd say if it's not needed for now let's remove it. We can always refer to the commit history if we want to add it back some day. Because plans change and we don't know for sure we're gonna add it back. If not, we'll end up with unused code and it's hard to maintain over time.

src/pages/Validators/Table.tsx
jillxuu: we didn't change anything to non-delegation nodes. we add a tab on top to choose from. you can see the screen recording video, which should give a clear cut on what has changed from this pr.

src/pages/Validators/ValidatorsTable.tsx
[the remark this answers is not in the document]
jillxuu: that makes sense, its just its a bit difficult to track if we add some of the data not all the data. right now i follow figma as the source of truth for all the info we need to show for delegation. but for the simplicity and clean code, i will remove it from this pr.

src/pages/Validators/Table.tsx
0xZihan: Sorry i should've been more clear. I meant since we're not changing any existing tables. We probably want to just keep the code as it was. And simply create a new file/component that is dedicated for delegation nodes table. Let me know if it makes sense to you.

jillxuu: after digging into the codebase to implement the separation of logic, it's as easy as we would have envisioned, cuz logics are really intertwined, delegation node and all nodes share certain columns. the current structure we have in `table.tsx` or `validatorsTable.tsx` make it very difficult to separate delegation specific logics out. even if we do so, it will result in code duplications. i would argue that having same file containing both delegation and all node logics make sense based on the current codebase structure, especially its still the same table but show different type of datas with tab change. lmk wyt! @0xZihan

0xZihan: There will be a couple of duplicated columns between `DelegatiionValidatorsTable` and `ValidatorsTable`, but rarely between `DelegatiionValidatorsTable` and `Table`. If you think about it, they're essentially just 3 different tables, with 3 different data sources. I don't think it makes the best sense to stick them together. 

That being said, I do agree that we should minimum duplicated code. We can do it by modularizing the shared components (columns). 

If you're blocked by this PR feel free to ship it. We can do refactor later.

jillxuu:
new revision with updated screen recording: 
https://user-images.githubusercontent.com/121921928/214450340-40acf7b1-44a3-44c8-8acb-5666a4254c9f.mov

src/pages/Validators/Index.tsx
```diff
@@ -1,7 +1,9 @@
-import {Box, Tab, Tabs, Typography} from "@mui/material";
+import {Box, ToggleButton, ToggleButtonGroup, Typography} from "@mui/material";
```
0xZihan, line 1: not needed

src/pages/Validators/Index.tsx
```diff
@@ -20,38 +22,54 @@ export default function ValidatorsPage() {
...
-            </Tabs>
-          ) : null}
-          <Box sx={{width: "auto", overflowX: "auto"}}>
-            <OldValidatorsTable onDelegatory={onDelegatory} />
-          </Box>
-        </>
-      )}
+      {state.network_name === "mainnet" && <ValidatorsMap />}
+      {inDev && validatorsTabs}
+      <Box sx={{width: "auto", overflowX: "auto"}}>{validatorsTable}</Box>
```
0xZihan, line 72: this looks like a long line. i could be wrong, but have we run `yarn fmt -w` to format the code?

0xZihan (approved):

src/pages/Validators/Index.tsx
jillxuu: haha this's after running the formatter. this line of code's prob right around the corner of the permitted line length.

PR #315, opened 2023-01-24
[Nerwork Select] persistent url param

0xZihan:
## Problem
The current network switching behaviour of explorer is a bit confusing. There is a URL parameter `?network=testnet` which designates the chosen network, but if one clicks a link that omits that it opens the explorer in whatever network the user had opened previously. This leads to some confusing link flows, esp when one opens links from different sources and may have mainnet and testnet explorer open in different tabs.

## Solution
Per @perryjrandall's suggestion, this PR makes it default to always include the `network` param in the url.

### DIFF
--- a/src/api/hooks/useNavigateWithParams.ts
+++ b/src/api/hooks/useNavigateWithParams.ts
@@ -1,0 +1,25 @@
+import {
+  createSearchParams,
+  generatePath,
+  useNavigate,
+  useSearchParams,
+} from "react-router-dom";
+import {defaultNetworkName} from "../../constants";
+
+export function useNavigateWithParams() {
+  const navigate = useNavigate();
+  const [searchParams] = useSearchParams();
+
+  const networkName = searchParams.get("network") || defaultNetworkName;
+
+  const navigateWithParams = (url: string) => {
+    const params = {network: networkName};
+    const path = generatePath(":url?:queryString", {
+      url,
+      queryString: createSearchParams(params).toString(),
+    });
+    navigate(path);
+  };
+
+  return navigateWithParams;
+}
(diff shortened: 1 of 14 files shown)


src/api/hooks/useNavigateWithParams.ts
```diff
@@ -0,0 +1,25 @@
...
+  useNavigate,
+  useSearchParams,
+} from "react-router-dom";
+import {defaultNetworkName} from "../../constants";
+
+export function useNavigateWithParams() {
+  const navigate = useNavigate();
+  const [searchParams] = useSearchParams();
+
+  const networkName = searchParams.get("network") || defaultNetworkName;
```
0xmaayan, line 13: what happens if url has `?network=zihan`?

0xmaayan (approved):

src/api/hooks/useNavigateWithParams.ts
0xZihan: it'll bring us to `?network=mainnet`

src/api/hooks/useNavigateWithParams.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  createSearchParams,
+  generatePath,
+  useNavigate,
+  useSearchParams,
+} from "react-router-dom";
+import {defaultNetworkName} from "../../constants";
+
+export function useNavigateWithParams() {
```
0xZihan, line 9: cc @jillxuu let's use `useNavigateWithParams` for future page navigation. This will ensure persistent network

0xZihan:
Actually this solution does not work well with `react-router-dom`. I'll close this out and work on other solutions.

perryjrandall:
Aww ok, excited to see other solution

Thanks for working on this!

PR #316, opened 2023-01-24
[explorer][delegation] refactor stake dialog UI

jillxuu:
Screenshot 2023-01-24 at 8 40 48 PM

https://user-images.githubusercontent.com/121921928/214438463-df05d70e-9b4b-4c54-bb9b-4419b56b8b5f.mov

### DIFF
--- a/src/components/IndividualPageContent/ContentRow.tsx
+++ b/src/components/IndividualPageContent/ContentRow.tsx
@@ -18,7 +18,7 @@ export default function ContentRow({
   value,
   tooltip,
   i,
-  container,
+  container = true,
   sx,
 }: ContentRowProps) {
   return (
--- a/src/components/IndividualPageContent/ContentRow.tsx
+++ b/src/components/IndividualPageContent/ContentRow.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import {Box, SxProps, Theme} from "@mui/material";
+import {Box} from "@mui/material";
 import Grid from "@mui/material/Unstable_Grid2";
 import {grey} from "../../themes/colors/aptosColorPalette";
 import EmptyValue from "./ContentValue/EmptyValue";
@@ -9,8 +9,6 @@ type ContentRowProps = {
   value: React.ReactNode;
   tooltip?: React.ReactNode;
   i?: any;
-  container?: boolean;
-  sx?: SxProps<Theme>;
 };
 
 export default function ContentRow({
@@ -18,18 +16,15 @@ export default function ContentRow({
   value,
   tooltip,
   i,
-  container = true,
-  sx,
 }: ContentRowProps) {
   return (
     <Box>
       <Grid
-        container={container && true}
+        container
         rowSpacing={0.5}
         columnSpacing={4}
         alignItems="start"
         key={i}
-        sx={sx}
       >
         <Grid xs={12} sm={3}>
           <Box sx={{fontSize: "0.875rem", color: grey[450]}}>
--- a/src/pages/DelegatoryValidator/DetailCard.tsx
+++ b/src/pages/DelegatoryValidator/DetailCard.tsx
@@ -1,8 +1,8 @@
-import {Box, Skeleton} from "@mui/material";
+import {Box, Skeleton, Stack} from "@mui/material";
 import * as React from "react";
 import HashButton from "../../components/HashButton";
 import ContentBox from "../../components/IndividualPageContent/ContentBox";
-import ContentRow from "../../components/IndividualPageContent/ContentRow";
+import ContentRowSpaceBetween from "../../components/IndividualPageContent/ContentRowSpaceBetween";
 import {HashType} from "../../components/HashButton";
 import RewardsPerformanceTooltip from "../Validators/Components/RewardsPerformanceTooltip";
 import LastEpochPerformanceTooltip from "../Validators/Components/LastEpochPerformanceTooltip";
@@ -41,12 +41,13 @@ export default function ValidatorDetailCard({
     }
   }, [lockedUntilSecs, operatorAddr, rewardGrowth, stakePoolAddress]);
 
+  // TODO: revisit layout for mobile
   return isSkeletonLoading ? (
     validatorDetailCardSkeleton()
   ) : (
-    <Box display="flex">
-      <ContentBox padding={4} width="50%" marginRight={3}>
-        <ContentRow
+    <Stack direction="row" spacing={4}>
+      <ContentBox width="50%" marginTop={0}>
+        <ContentRowSpaceBetween
           title={"Operator"}
           value={
             operatorAddr && (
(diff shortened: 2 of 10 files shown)


src/components/IndividualPageContent/ContentRow.tsx
```diff
@@ -18,7 +18,7 @@ export default function ContentRow({
   value,
   tooltip,
   i,
-  container,
+  container = true,
```
jillxuu, line 21: this is to set the container boolean default to true so that if no `container` pass in, it will still have `container={true}`. goal is to fix the current misalignment on the UI, so that it will be inline instead of in two lines.
Screenshot 2023-01-24 at 2 51 13 PM

src/pages/DelegatoryValidator/DetailCard.tsx
```diff
@@ -76,18 +88,21 @@ export default function ValidatorDetailCard({
...
         />
         <ContentRow
+          container={false}
           sx={sx}
           title="Last Epoch Performance"
           value={validator ? validator.last_epoch_performance : null}
           tooltip={<LastEpochPerformanceTooltip />}
         />
         <ContentRow
+          container={false}
```
jillxuu, line 105: for the details card below, we don't want container effect but rather have `space-between` effect: 
Screenshot 2023-01-24 at 2 52 42 PM

src/pages/DelegatoryValidator/StakingBar.tsx
```diff
@@ -12,12 +12,17 @@ import ContentBox from "../../components/IndividualPageContent/ContentBox";
 import {getFormattedBalanceStr} from "../../components/IndividualPageContent/ContentValue/CurrencyValue";
 import StakeDialog from "./StakeDialog";
 import ArrowCircleUpIcon from "@mui/icons-material/ArrowCircleUp";
+import {Types} from "aptos";
 
 type ValidatorStakingBarProps = {
   validator: MainnetValidatorData;
+  accountResource?: Types.MoveResource | undefined;
```
jillxuu, line 19: decide to pass in `accountResource` even though right now we only need unlock time, in the future when more data sources coming in, we might need `accountResource` in other use cases.

src/pages/DelegatoryValidator/DetailCard.tsx
```diff
@@ -62,11 +63,22 @@ export default function ValidatorDetailCard({
           value={null}
           sx={sx}
         />
-        <ContentRow title="Compound Rewards" value={null} sx={sx} />
-        <ContentRow title="Operator Commission" value={null} sx={sx} />
+        <ContentRow
```
0xmaayan, line 66: looks like `container` is an optional prop, you might simply dont need to pass it here - so it would be false/undefined in the `ContentRow` component

src/pages/DelegatoryValidator/StakingBar.tsx
```diff
@@ -81,7 +86,11 @@ export default function StakingBar({validator}: ValidatorStakingBarProps) {
           <Typography>Stake</Typography>
         </Button>
       </ListItem>
-      <StakeDialog handleDialogClose={handleClose} isDialogOpen={dialogOpen} />
+      <StakeDialog
+        handleDialogClose={handleClose}
+        isDialogOpen={dialogOpen}
```
0xmaayan, line 91: is the Dialog being used anywhere else on this file? if not, probably can more all Dialog related stuff to the `StakeDialog` file

src/pages/DelegatoryValidator/StakeDialog.tsx
```diff
@@ -1,51 +1,134 @@
...
+              <InputAdornment
+                position="end"
+                sx={{ml: 0.5, marginTop: "0!important"}}
+              >
+                APT
+              </InputAdornment>
+            }
+          />
+          {stakedAmountError && <FormHelperText error>Error</FormHelperText>}
+        </DialogContent>
```
0xmaayan, line 93: not completely sure how MUI Dialog works, but I would assume it should have a  one `<DialogContent >` component..?

```
<DialogTitle />
   <DialogContent >
      ... content ....
   </DialogContent >
<DialogActions />
```

src/pages/DelegatoryValidator/StakingBar.tsx
jillxuu: we will also have a `unstakedialog` and `transaction succeed` dialog. in the future will consolidate all the dialog UI to a single component file.

src/pages/DelegatoryValidator/StakeDialog.tsx
jillxuu: you were referring to have this instead?
```
<DialogContent >
      ... single content ....
   </DialogContent >
<DialogContent >
      ... single content ....
   </DialogContent >
```
the built-in css layouts for `DialogContent` is a bit too much padding/margin for us, that's why i keep all the content into a single `DialogContent`

0xmaayan: looks like there is 4 different `DialogContent` components inside the parent `Dialog` component, is that how it should be..?

src/pages/DelegatoryValidator/DetailCard.tsx
jillxuu: the current way of making it work is:
1. container is an optional boolean variable
1.1 if container's not specified, default to `true`, so that `container={container && true}` => `container={true}`.
1.2 if container's specified as `false`, so that `container={container && true}` => `container={false}`. which is what we want in `DetailCard` cuz `container={true}` will not make "flex and space-between" work.
1.3 if container's specified as `true`, same to 1.1

so the example above, if we don't pass in `container={false}`, `container` will equal to `true` in `ContentRow`, not the exp we want. 

so you may ask, why not the other way around. set it default to false, and don't pass in container variable. it will work to this specific use case, but then we need to change all the code pointers where `ContentRow` is used to have then pass in `container={true}` instead.

a bit complicated, lmk if you have any questions!

src/pages/DelegatoryValidator/StakeDialog.tsx
jillxuu: good call. by looking at the MUI site: https://mui.com/material-ui/react-dialog/, `DialogContent` is used to wrap about the content. will remove some of them and use `Stack` for spacing around each block.

src/pages/DelegatoryValidator/StakeDialog.tsx
```diff
@@ -5,47 +5,113 @@ import {
...
-        epoch (~2 hours) due to the delay time
-      </DialogContent>
+    <Dialog open={isDialogOpen} onClose={handleDialogClose}>
+      <IconButton
+        onClick={handleDialogClose}
+        sx={{
+          position: "absolute",
+          right: 12,
+          top: 12,
+          color: (theme) => theme.palette.grey[500],
```
0xZihan, line 55: if you ever try to find a "middle" color that works for both dark and light mode, use `grey[450]`

src/pages/DelegatoryValidator/StakeDialog.tsx
```diff
@@ -5,47 +5,113 @@ import {
...
-      </ContentBox>
-      <DialogActions>
-        <Button onClick={handleDialogClose}>Deposit</Button>
-      </DialogActions>
-      <DialogContent>
-        Be aware that you will be able to see your funds in the pool after 1
-        epoch (~2 hours) due to the delay time
-      </DialogContent>
+    <Dialog open={isDialogOpen} onClose={handleDialogClose}>
+      <IconButton
```
0xZihan, line 49: it's a BE work so we can do it later but it's worth it to create a `StyledDialog` component with this "X" icon in it and get it reused for whenever we need a dialog/modal. For example, wallet button modal should have the same style.

0xZihan (approved):

PR #326, opened 2023-01-27
[explorer][delegation] add status icon to my deposits section

jillxuu:
share this step status across tooltip and table row. 

Screenshot 2023-01-27 at 11 36 53 AM


0xZihan:
I would argue a better way to do this is to create a component for the this and reuse the component in both tooltip and column.
Screen Shot 2023-01-27 at 12 56 41 PM

0xmaayan (approved):

PR #330, opened 2023-01-30
[explorer][hot-fix] validators page geoData null issue

jillxuu:
Screenshot 2023-01-30 at 11 20 06 AM

validators page fail to load due to country potentially being null for this validator
Screenshot 2023-01-30 at 11 20 48 AM

### DIFF
--- a/src/api/hooks/useGetMainnetValidatorsGeoData.ts
+++ b/src/api/hooks/useGetMainnetValidatorsGeoData.ts
@@ -40,7 +40,7 @@ export function useGetValidatorSetGeoData() {
     const groups: ValidatorGeoGroup[] = validators.reduce(
       (groups: ValidatorGeoGroup[], validatorData: MainnetValidatorData) => {
         const geoData = validatorData.location_stats;
-        const country = geoData.country;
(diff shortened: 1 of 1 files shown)


0xmaayan (approved):

src/api/hooks/useGetMainnetValidatorsGeoData.ts
```diff
@@ -40,7 +40,7 @@ export function useGetValidatorSetGeoData() {
     const groups: ValidatorGeoGroup[] = validators.reduce(
       (groups: ValidatorGeoGroup[], validatorData: MainnetValidatorData) => {
         const geoData = validatorData.location_stats;
-        const country = geoData.country;
+        const country = geoData?.country;
```
zacharydenton, line 43: I suggest changing this type https://github.com/aptos-labs/explorer/blob/61cd024c7e4bfd01a452abd1badfa10420f0b606/src/api/hooks/useGetMainnetValidators.ts#L17 to reflect the fact that it can be `undefined`.

zacharydenton (approved):

PR #335, opened 2023-01-31
[explorer][delegation] address some nits

jillxuu:
https://user-images.githubusercontent.com/121921928/215843907-763aa198-e369-450d-88fe-7c06a22c04ac.mov

couple things this pr addressed:

1. on the validators page, the tab has underline colors previously. pr updated to get rid of the color and remain consistent with tab usages in other places
2. lockedUtilSecs could be nullable. check it whether it's null or not before hands. this eliminates the error in the react console.
3.  make details page skeleton mobile friendly
4. address react console warning on dialog font size complain: "h5 can't be in h2"
5. short term sol:  persist the dev param in the url from the validators page to the validator page

### DIFF
--- a/src/pages/DelegatoryValidator/WalletConnectionDialog.tsx
+++ b/src/pages/DelegatoryValidator/WalletConnectionDialog.tsx
@@ -29,10 +29,8 @@ export default function WalletConnectionDialog({
       <Box sx={{display: "flex", justifyContent: "center", marginY: 2}}>
         <ConnectWalletModalIcon />
       </Box>
-      <DialogTitle>
-        <Typography variant="h5" textAlign="center">
-          Please connect your wallet
-        </Typography>
+      <DialogTitle textAlign="center">
+        <div>Please connect your wallet</div>
         <Typography variant="caption" textAlign="center">
           You need to connect your wallet to be able to stake
         </Typography>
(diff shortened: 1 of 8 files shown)


src/pages/DelegatoryValidator/WalletConnectionDialog.tsx
```diff
@@ -29,10 +29,8 @@ export default function WalletConnectionDialog({
       <Box sx={{display: "flex", justifyContent: "center", marginY: 2}}>
         <ConnectWalletModalIcon />
       </Box>
-      <DialogTitle>
-        <Typography variant="h5" textAlign="center">
-          Please connect your wallet
-        </Typography>
+      <DialogTitle textAlign="center">
```
0xmaayan, line 32: do we also want to have `variant="h5"` here?

src/pages/Validators/ValidatorsTable.tsx
```diff
@@ -321,9 +321,17 @@ type ValidatorRowProps = {
 function ValidatorRow({validator, columns}: ValidatorRowProps) {
   const inDev = useGetInDevMode();
   const navigate = useNavigate();
+  const [searchParams] = useSearchParams();
 
   const rowClick = (address: Types.Address) => {
-    navigate(`/validator/${address}`);
+    // TODO(jill) find long term to persist the url params
+    const isUsingDev = searchParams.get("feature");
```
0xmaayan, line 328: there should be a hook somewhere `isDevMode` (or something like that), why not using it?

0xmaayan (approved):

src/pages/DelegatoryValidator/WalletConnectionDialog.tsx
jillxuu: this DialogTitle includes two texts, not both of them need h5 size here.

PR #339, opened 2023-02-02
[explorer][delegation] refresh time countdown interval bar every min

jillxuu:
this screen recording is a bit long, feel free to fast forward to see the time change without refreshing the page. 

https://user-images.githubusercontent.com/121921928/216201765-af4734e1-012d-4679-88c4-92e6b65a29d0.mov

### DIFF
--- a/src/pages/DelegatoryValidator/Components/TimeDurationIntervalBar.tsx
+++ b/src/pages/DelegatoryValidator/Components/TimeDurationIntervalBar.tsx
@@ -3,6 +3,8 @@ import React, {useEffect, useState} from "react";
 import IntervalBar from "../../../components/IntervalBar";
 import {parseTimestamp, timestampDisplay} from "../../utils";
 
+const REFRESH_IN_MS = 60000;
+
 export default function TimeDurationIntervalBar({
   timestamp,
 }: {
@@ -22,7 +24,7 @@ export default function TimeDurationIntervalBar({
   const unlockTime = parseTimestamp(timestamp.toString());
 
   useEffect(() => {
-    const refresh = () => {
+    const interval = setInterval(() => {
       // the remaining time of the unlock cycle
       const now = moment();
       const remainingTime = moment.duration(
@@ -42,16 +44,16 @@ export default function TimeDurationIntervalBar({
           (unlockTime.valueOf() - startTime.valueOf())) *
         100;
       setPercentageComplete(percentage);
-    };
-
-    refresh();
-    setInterval(refresh, 60 * 1000); // refresh every minute
+    }, REFRESH_IN_MS);
 
     const remainingTimeInMS = remainingTime?.asMilliseconds();
 
     if (remainingTimeInMS !== undefined && remainingTimeInMS <= 0) {
       window.location.reload();
     }
+
+    // This represents the unmount function, in which you need to clear your interval to prevent memory leaks.
+    return () => clearInterval(interval);
   }, [remainingTime, percentageComplete]);
 
   const remainingTimeDisplay = timestampDisplay(
--- a/src/pages/DelegatoryValidator/Components/TimeDurationIntervalBar.tsx
+++ b/src/pages/DelegatoryValidator/Components/TimeDurationIntervalBar.tsx
@@ -1,5 +1,5 @@
 import moment from "moment";
-import React from "react";
+import React, {useEffect, useState} from "react";
 import IntervalBar from "../../../components/IntervalBar";
 import {parseTimestamp, timestampDisplay} from "../../utils";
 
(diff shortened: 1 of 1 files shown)


src/pages/DelegatoryValidator/Components/TimeDurationIntervalBar.tsx
```diff
@@ -12,37 +12,55 @@ export default function TimeDurationIntervalBar({
...
+
+    refresh();
+    setInterval(refresh, 60 * 1000); // refresh every minute
+
+    const remainingTimeInMS = remainingTime?.asMilliseconds();
+
+    if (remainingTimeInMS !== undefined && remainingTimeInMS <= 0) {
+      window.location.reload();
+    }
+  }, [remainingTime, percentageComplete]);
```
0xmaayan, line 55: do we want to update a component every minute? I think we can simply do
```
const MINUTE_MS = 60000;

useEffect(() => {
  const interval = setInterval(() => {
    console.log('Logs every minute');
  }, MINUTE_MS);

  return () => clearInterval(interval); // This represents the unmount function, in which you need to clear your interval to prevent memory leaks.
}, [])
```

src/pages/DelegatoryValidator/Components/TimeDurationIntervalBar.tsx
```diff
@@ -12,37 +14,55 @@ export default function TimeDurationIntervalBar({
...
+
+    const remainingTimeInMS = remainingTime?.asMilliseconds();
+
+    if (remainingTimeInMS !== undefined && remainingTimeInMS <= 0) {
+      window.location.reload();
+    }
+
+    // This represents the unmount function, in which you need to clear your interval to prevent memory leaks.
+    return () => clearInterval(interval);
+  }, [remainingTime, percentageComplete]);
```
0xmaayan, line 57: not completely understand why you use `remainingTime` and `percentageComplete` as a `useEffect` dependency ..? that means that this hook would be called any time the state is changes

jillxuu: great catch! the dependencies i have here caused max stack depth rendering, cuz state change will trigger `useEffect` and useEffect will change state, endless loop.

src/pages/DelegatoryValidator/Components/TimeDurationIntervalBar.tsx
```diff
@@ -12,37 +14,55 @@ export default function TimeDurationIntervalBar({
...
+      const percentage =
+        (alreadyPassedTime.asMilliseconds() /
+          (unlockTime.valueOf() - startTime.valueOf())) *
+        100;
+      setPercentageComplete(percentage);
+
+      const remainingTimeInMS = remainingTime?.asMilliseconds();
+
+      if (remainingTimeInMS !== undefined && remainingTimeInMS <= 0) {
+        window.location.reload();
```
0xmaayan, line 51: why reload? 

sorry for all this partial review, seems like I keep missing things =)

src/pages/DelegatoryValidator/Components/TimeDurationIntervalBar.tsx
```diff
@@ -12,37 +14,55 @@ export default function TimeDurationIntervalBar({
...
   // the end of the unlock cycle
   const unlockTime = parseTimestamp(timestamp.toString());
 
-  // the remaining time of the unlock cycle
-  const now = moment();
-  const remainingTime = moment.duration(
-    unlockTime.valueOf() - now.valueOf(),
-    "milliseconds",
-  );
+  useEffect(() => {
```
0xmaayan, line 26: btw, we do something similar on governance https://github.com/aptos-labs/governance/blob/main/src/pages/Proposal/Header.tsx#L73-L85 not sure if its exactly this use case but might be good to check it out

src/pages/DelegatoryValidator/Components/TimeDurationIntervalBar.tsx
jillxuu: we had reports on the validators page's epoch interval that it passes 100% and time remaining becomes negative if we don't refresh the page and its over 2 hours epoch time. seems like time interval we used [epoch implementation](https://github.com/aptos-labs/explorer/blob/20db87898930bcba5e7de32b53ebc2ac8cd512e1/src/pages/Validators/Components/Epoch.tsx#L20) will keep updating every min but not aware of the boundary. so i decide to use window reload to force reloading the window when time's up so that we don't see negative time countdown.

Screenshot 2023-02-02 at 5 06 55 PM

0xmaayan: is there anything specific we need to show when it passes 100%? because we can
```
if (remainingTimeInMS !== undefined && remainingTimeInMS <= 0) {
      // update component to display something we want
      clearInterval(interval);
}
```

jillxuu: when it passes 100%, it will reset to another 30 days countdown (30 day is the staking cycle). so the ideal situation would be: when remaining time reaching 0, we keep computing the remaining time with the presumption that unlock time will be updated to 30 days from now. are you suggesting we re-compute remainingTime in the code block like:

```
if (remainingTimeInMS !== undefined && remainingTimeInMS <= 0) {
      const remainingTime = moment.duration(
        unlockTime.valueOf() - now.valueOf(),
        "milliseconds",
      );
      clearInterval(interval);
}
```

alternatively, if we don't do anything in the code block, but just clear the interval, like this [code pointer](https://github.com/aptos-labs/explorer/blob/20db87898930bcba5e7de32b53ebc2ac8cd512e1/src/pages/Validators/Components/Epoch.tsx#L53-L55) is doing, this is why people had issues like [WP-289](https://linear.app/aptoslabs/issue/WP-289/epoch-progress-bar-goes-over-100percent-and-time-remaining-becomes)?

src/pages/DelegatoryValidator/Components/TimeDurationIntervalBar.tsx
```diff
@@ -12,37 +14,66 @@ export default function TimeDurationIntervalBar({
...
+    const refresh = () => {
```
jillxuu, line 27: i change it back to using refresh cuz i realized with directly using the `setInterval` the first time it will not be ready but we need to wait like 1 min afterwards for it to take effect. creating a function called `refresh` and call it right after would trigger it right away. as suggested in this [post](https://stackoverflow.com/questions/6685396/execute-the-setinterval-function-without-delay-the-first-time).

0xmaayan (approved):

PR #343, opened 2023-02-06
[explorer] ctrl+f or cmd+f search validator pool address WP-298

jillxuu:
https://user-images.githubusercontent.com/121921928/217100229-12e89532-e070-400a-891c-607e5a279b31.mov

### DIFF
--- a/src/components/HashButton.tsx
+++ b/src/components/HashButton.tsx
@@ -60,20 +60,36 @@ interface HashButtonProps extends BoxProps {
   hash: string;
   type: HashType;
   size?: "small" | "large";
+  shouldTruncate?: boolean;
 }
 
 export default function HashButton({
   hash,
   type,
   size = "small",
+  shouldTruncate = true,
   ...props
 }: HashButtonProps) {
   if (type === HashType.ACCOUNT) {
     return (
-      <AccountHashButtonInner hash={hash} type={type} size={size} {...props} />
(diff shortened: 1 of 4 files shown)


src/components/HashButton.tsx
```diff
@@ -132,8 +153,12 @@ function HashButtonInner({
   const open = Boolean(anchorEl);
   const id = open ? "simple-popover" : undefined;
 
-  const truncateHash =
-    size === "large" ? truncateAddressMiddle(hash) : truncateAddress(hash);
+  // detect key event search to not truncate the hash so that search is easier
+  const truncateHash = !shouldTruncate
```
0xZihan, line 157: Usually nested ternary is not encouraged for readability purpose.

0xZihan (changes requested):
Thanks for the swift work!! It's a smart solution to extend the hashes upon searching. Although I'm not sure if it is the best choice from design perspective. For example, when we exit cmd+f search bar, the hashes are not collapsed. And visually the table is quite wide when it's extended. Also I'm wondering if people would want to search for operator address or name. So I would suggest let's have this discussion with our designer and explore a better way to solve this. I can imagine having a search bar for tables like this. Or is there a way that we "hide" the addresses but make it searchable. Let's explore the options. 

Right now to kinda unblock the users, they can search with the first few of the last few characters in the address string.

PR #344, opened 2023-02-08
[explorer][delegation] add banner for new feature launch

jillxuu:
https://user-images.githubusercontent.com/121921928/217400691-08b4272b-63e9-4cf3-b216-28c43ae81dd8.mov

### DIFF
--- a/src/pages/layout/index.tsx
+++ b/src/pages/layout/index.tsx
@@ -56,15 +56,7 @@ export default function ExplorerLayout({children}: LayoutProps) {
                       theme.palette.mode === "dark" ? "#1E40AF" : "#1D4ED8",
                     borderRadius: 0,
                   }}
-                  icon={
-                    <Stack color="#ffffff">
-                      <link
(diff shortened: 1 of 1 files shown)


src/pages/layout/index.tsx
```diff
@@ -29,6 +48,57 @@ export default function ExplorerLayout({children}: LayoutProps) {
...
+                        }}
+                      >
+                        <CloseIcon fontSize="inherit" />
+                      </IconButton>
+                    </Stack>
+                  }
+                >
+                  <Typography color="#ffffff">
+                    Power Aptos network. Stake tokens and earn rewards.
+                  </Typography>
```
0xZihan, line 98: Is this centered? Feels a bit higher or it's just my eyes.

0xZihan (approved):
Stay tuned that we might have some change coming up to the design. One thing is we're taking off the icon. Maybe you want to remove it here too.

src/pages/layout/index.tsx
jillxuu: its prob cuz of the icon. once we remove it, its in the center now

PR #345, opened 2023-02-10
[explorer][delegation] add view function query to get stake info per delegator

jillxuu:
### right now using testnet delegation pool address manually

Screenshot 2023-02-09 at 5 16 43 PM
Screenshot 2023-02-09 at 5 16 51 PM

### DIFF
--- a/src/api/hooks/useGetDelegatorStakeInfo.ts
+++ b/src/api/hooks/useGetDelegatorStakeInfo.ts
@@ -9,15 +9,12 @@ export function useGetDelegatorStakeInfo(delegatorAddress: Types.Address) {
   const [stake, setStake] = useState<Types.MoveValue[]>([]);
 
   useEffect(() => {
-    if (stake !== undefined) {
-      const fetchData = async () => {
-        const stakes = await getStake(state.network_value, delegatorAddress);
-        setStake(stakes);
-        setIsLoading(false);
-      };
-      fetchData();
-    }
-  }, [stake, state]);
+    const fetchData = async () => {
+      setStake(await getStake(state.network_value, delegatorAddress));
+      setIsLoading(false);
+    };
+    fetchData();
+  }, [state]);
 
   return {stake, isLoading};
 }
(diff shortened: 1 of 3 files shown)


src/api/index.ts
```diff
@@ -187,3 +187,21 @@ export async function getRecentBlocks(
...
+
+// TODO(jill): update this function name once in mainnet
+export async function getStake(
+  nodeUrl: string,
+  delegatorAddress: Types.Address,
+): Promise<Types.MoveValue[]> {
+  const client = new AptosClient(nodeUrl);
+  const payload: Types.ViewRequest = {
+    function:
+      "0x1310dc820487f24755e6e06747f6582118597a48868e2a98260fa8c3ee945cbd::delegation_pool::get_stake",
```
movekevin, line 199: Can we define the address here as a constant somewhere? This would allow changing it to 0x1 more easily later on.

src/api/index.ts
```diff
@@ -187,3 +188,20 @@ export async function getRecentBlocks(
   }
   return blocks;
 }
+
+export async function getStake(
+  nodeUrl: string,
+  delegatorAddress: Types.Address,
+): Promise<Types.MoveValue[]> {
+  const client = new AptosClient(nodeUrl);
```
0xmaayan, line 196: not sure we want create a new instance every time we call this function (which seems like it is every time stake or state is changing) - ideally we would have one `client` instance in our app that changes when the `network` changes (maybe a good place for it is in global state).

Sounds it is out of this PR scope, but something to look at

src/api/hooks/useGetDelegatorStakeInfo.ts
```diff
@@ -0,0 +1,23 @@
...
+export function useGetDelegatorStakeInfo(delegatorAddress: Types.Address) {
+  const [state, _] = useGlobalState();
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [stake, setStake] = useState<Types.MoveValue[]>([]);
+
+  useEffect(() => {
+    if (stake !== undefined) {
+      const fetchData = async () => {
+        const stakes = await getStake(state.network_value, delegatorAddress);
+        setStake(stakes);
```
0xmaayan, line 15: a bit confused here, looks like when `stake` is defined, we get the stake? and then set the stake state, which triggers the useEffect again and start this whole cycle?

src/api/index.ts
jillxuu: WP-325 linear issue created!

0xZihan (approved):

PR #347, opened 2023-02-14
[explorer][delegation] add operator commission percentage from sc view function

jillxuu:
https://user-images.githubusercontent.com/121921928/218859803-3030463b-10f1-4811-a2d2-229b1defe298.mov

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -207,9 +207,8 @@ export async function getStake(
 }
 
 export async function getValidatorCommission(
-  nodeUrl: string,
+  client: AptosClient,
 ): Promise<Types.MoveValue> {
-  const client = new AptosClient(nodeUrl);
   const payload: Types.ViewRequest = {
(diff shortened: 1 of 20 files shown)


src/api/index.ts
```diff
@@ -205,3 +205,18 @@ export async function getStake(
   };
   return await client.view(payload);
 }
+
+export async function getValidatorCommission(
+  nodeUrl: string,
+): Promise<Types.MoveValue> {
+  const client = new AptosClient(nodeUrl);
```
jillxuu, line 212: i add aptos client in the global state in a previous pr but hooks can't be used in regular functions, it has to be either custom hooks or function components. any suggestions on how to abstract out aptos client here? @0xmaayan

jillxuu (commented):
i have this branch on top of another development branch but previous pr branch's commit is also here, unexpected! followed the github branch instruction, not sure which part i did wrong. ideally this pr should only show its branch's change but this branch's change's based on top of another branch.

PR #353, opened 2023-02-21
[explorer] update the primary color for explorer

jillxuu:
# change
I notice that in explorer, we have different primaryColorToned. We used `primary[600]` but ANS uses `primary[500]`. to keep the consistency, i believe it makes sense to keep both of them as `primary[500]` so that all the components that's primary will show consistent color palette. 

| before | after |
| -- | -- |
| Screenshot 2023-02-20 at 5 48 34 PM | Screenshot 2023-02-20 at 5 48 16 PM |

### DIFF
--- a/src/themes/theme.ts
+++ b/src/themes/theme.ts
@@ -65,7 +65,7 @@ declare module "@mui/material/Divider" {
 }
 
 const primaryColor = primary["400"];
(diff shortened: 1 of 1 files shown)


0xZihan:
I'll leave this one for @moonclavedev

PR #355, opened 2023-02-21
[explorer][delegation] only enable validator table row click if its whitelisted delegation nodes

jillxuu:
https://user-images.githubusercontent.com/121921928/220244812-6a99ecd5-ad98-4dcc-8357-cdea6e55a509.mov

### DIFF
--- a/src/pages/Validators/ValidatorsTable.tsx
+++ b/src/pages/Validators/ValidatorsTable.tsx
@@ -378,7 +378,12 @@ function ValidatorRow({validator, columns}: ValidatorRowProps) {
 
   return (
     <GeneralTableRow
-      onClick={inDev ? () => rowClick(validator.owner_address) : undefined}
+      onClick={
+        inDev &&
+        WHILTELISTED_TESTNET_DELEGATION_NODES.includes(validator.owner_address)
+          ? () => rowClick(validator.owner_address)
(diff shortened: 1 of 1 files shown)


src/pages/Validators/ValidatorsTable.tsx
```diff
@@ -378,7 +378,12 @@ function ValidatorRow({validator, columns}: ValidatorRowProps) {
 
   return (
     <GeneralTableRow
-      onClick={inDev ? () => rowClick(validator.owner_address) : undefined}
+      onClick={
+        inDev &&
+        WHILTELISTED_TESTNET_DELEGATION_NODES.includes(validator.owner_address)
+          ? () => rowClick(validator.owner_address)
+          : undefined
+      }
```
0xZihan, line 386: this is why i think normal nodes and delegatory node should be two tables. there'll be more and more if/else checks. let's make sure we have a BE follow-up to address this.

0xZihan (approved):

src/pages/Validators/ValidatorsTable.tsx
jillxuu: yes! thats what i thought and here's the PR for the separation: https://github.com/aptos-labs/explorer/pull/358

PR #357, opened 2023-02-21
[explorer][delegation] handle bigint overflow when submitting txn

jillxuu:
stake amount is bigint, it can only take amount thats in the signed bigint range. however, it's easy to exceed that limit and overflow. right now, this pr used `toFixed(8)` to handle the limit but its not ideal.  open for suggestions on how to better handle this situation. 

### before 
https://user-images.githubusercontent.com/121921928/220262873-83a21638-b7a3-472f-92b2-3fa76381a7e2.mov

### after
https://user-images.githubusercontent.com/121921928/220262878-ce6aa345-3d26-4357-a755-b2d6aca20299.mov

### DIFF
--- a/src/pages/DelegatoryValidator/StakeOperationDialog.tsx
+++ b/src/pages/DelegatoryValidator/StakeOperationDialog.tsx
@@ -91,7 +91,7 @@ export default function StakeOperationDialog({
     if (isAmountValid) {
       await submitStakeOperation(
         validator.owner_address,
-        Number(Number(amount).toFixed(8)),
+        Number((Number(amount) * OCTA).toFixed(0)),
         stakeOperation,
       );
     }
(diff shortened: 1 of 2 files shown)


src/pages/DelegatoryValidator/StakeOperationDialog.tsx
```diff
@@ -92,7 +91,7 @@ export default function StakeOperationDialog({
     if (isAmountValid) {
       await submitStakeOperation(
         validator.owner_address,
-        Number(amount),
+        Number(Number(amount).toFixed(8)),
```
0xmaayan, line 94: can use BigInt()..?
https://github.com/aptos-labs/explorer/blob/main/src/api/index.ts#L35

jillxuu: i don't think it worked: if we just BigInt(a number), if that number exceeds the bigint allowed limit, it will still throw errors
Screenshot 2023-02-21 at 10 11 27 AM

0xZihan (approved):
can you fix to 8 and then convert to bigint

jillxuu:
> can you fix to 8 and then convert to bigint

SC's not taking bigint but number instead. keep using the number but toFixed(0) with OCTA.

PR #363, opened 2023-02-22
[explorer][delegation] fix delegatedStakeAmount

jillxuu:
before, we were using validator voting power for delegatedStakeAmount, which is wrong. now we get the right amount for staking from [account resources](https://fullnode.testnet.aptoslabs.com/v1/accounts/0x5df905f817adf39293c596e83512ab8a9dc5a19980e11bd4ce44b6e749d33a0d/resources) delegation pool's total coins. 
Screenshot 2023-02-22 at 10 42 17 AM

Screenshot 2023-02-22 at 10 42 10 AM

### DIFF
--- a/src/api/hooks/useGetDelegationNodeInfo.ts
+++ b/src/api/hooks/useGetDelegationNodeInfo.ts
@@ -1,14 +1,23 @@
 import {AptosClient, Types} from "aptos";
 import {useState, useMemo} from "react";
 import {getValidatorCommission} from "..";
+import {DELEGATION_POOL_ADDRESS} from "../../constants";
 import {useGlobalState} from "../../GlobalState";
+import {useGetAccountResources} from "./useGetAccountResources";
 import {ValidatorData} from "./useGetValidators";
 import {useGetValidatorSet, Validator} from "./useGetValidatorSet";
 
+interface DelegationPool {
+  active_shares: {
+    total_coins: string;
+  };
+}
+
 type DelegationNodeInfoResponse = {
-  delegatedStakeAmount: string;
+  delegatedStakeAmount: string | undefined;
   networkPercentage?: string;
   commission: number | undefined;
+  isLoading: boolean;
 };
 
 type DelegationNodeInfoProps = {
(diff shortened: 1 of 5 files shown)


src/api/hooks/useGetDelegationNodeInfo.ts
```diff
@@ -1,14 +1,23 @@
...
+    total_coins: string;
+  };
+}
+
 type DelegationNodeInfoResponse = {
-  delegatedStakeAmount: string;
+  delegatedStakeAmount: string | undefined;
   networkPercentage?: string;
   commission: number | undefined;
+  isLoading: boolean;
```
0xmaayan, line 20: I don't think `isLoading` should be a type property. If you want to use it to indicate that the state is `isLoading` we should hold a local state (useState), similar to what we are doing here
https://github.com/aptos-labs/governance/blob/main/src/api/hooks/useSubmitTransaction.ts#L28

src/api/hooks/useGetDelegationNodeInfo.ts
```diff
@@ -21,27 +30,46 @@ export function useGetDelegationNodeInfo({
   validator,
 }: DelegationNodeInfoProps): DelegationNodeInfoResponse {
   const [state, _] = useGlobalState();
-  const [commission, setCommission] = useState<Types.MoveValue>();
-  const client = new AptosClient(state.network_value);
   const {totalVotingPower} = useGetValidatorSet();
+  const {data, isLoading} = useGetAccountResources(validatorAddress);
```
0xmaayan, line 34: not completely sure where `isLoading` coming from? doesn't seem that it is returned from `useGetAccountResources`
https://github.com/aptos-labs/explorer/blob/main/src/api/hooks/useGetAccountResources.ts#L19

src/api/hooks/useGetDelegationNodeInfo.ts
```diff
@@ -21,27 +30,46 @@ export function useGetDelegationNodeInfo({
...
+      const delegationPool = data?.find(
+        (resource) =>
+          resource.type ===
+          `${DELEGATION_POOL_ADDRESS}::delegation_pool::DelegationPool`,
+      );
+
+      const delegationPoolData: DelegationPool | undefined =
+        delegationPool?.data as DelegationPool;
+
+      setDelegatedStakeAmount(delegationPoolData?.active_shares?.total_coins);
```
0xmaayan, line 63: looks like you set the [`total_coins` type as `string`](https://github.com/aptos-labs/explorer/pull/363/files#diff-ae78229edfd40cc88cd35326bd4731041f41fde373b63acacb9af87949c8526aR12), but set it here in a [state of type `Types.MoveValue`](https://github.com/aptos-labs/explorer/pull/363/files#diff-ae78229edfd40cc88cd35326bd4731041f41fde373b63acacb9af87949c8526aR38)

src/api/hooks/useGetDelegationNodeInfo.ts
0xmaayan, line 34: looks like you are looking for a specific resource - can use [useGetAccountResource](https://github.com/aptos-labs/explorer/blob/main/src/api/hooks/useGetAccountResource.ts)

jillxuu: haha, `isLoading` is one of the returns from `useQuery`. a better example would be [here](https://github.com/aptos-labs/explorer/blob/b88dcecf3756b99afe9ea2ba37c6db7d4efcf970/src/api/hooks/useGetAccountResource.ts#L28) 


```
const accountResourcesResult = useQuery<
    Array<Types.MoveResource>,
    ResponseError
  >(
    ["accountResource", {address}, state.network_value],
    () => getAccountResources({address}, state.network_value),
    {refetchOnWindowFocus: false},
  );
  
  const {isLoading, isError, refetch} = accountResourcesResult;
```

0xZihan (approved):

PR #368, opened 2023-02-23
[explorer][delegation] add node tabs to the URL

jillxuu:
### changes
1. add tab change to URL
2. make feature flag persistent when navigating around the website
3. remove devnet delegation info cuz its not needed

https://user-images.githubusercontent.com/121921928/221254080-39f35439-ea09-4ec3-8ae8-04d99df231f7.mov


0xZihan (changes requested):
Can we create a `Tabs.tsx` component and follow the structure in other pages (account, transaction, token, ...)? Please see https://github.com/aptos-labs/explorer/blob/main/src/pages/Account/Tabs.tsx for example.

0xZihan (approved):

PR #369, opened 2023-02-23
[explorer][delegation] add number of delegators from gql query

jillxuu:
https://user-images.githubusercontent.com/121921928/221255676-94fee9a3-5061-4150-88d8-8473fd1d6dc4.mov


src/api/hooks/useGetDelegationNodeInfo.ts
```diff
@@ -9,13 +10,59 @@ type DelegationNodeInfoResponse = {
...
+      distinct_on: delegator_address
+    ) {
+      aggregate {
+        count
+      }
+    }
+  }
+`;
+
+function useGetNumberOfDelegators(poolAddress: Types.Address) {
```
0xZihan, line 47: why not just export it a separate hook and use it directly? adding this as part of delegation info will introduce unnecessary dependency.

0xZihan (changes requested):

0xZihan (approved):

PR #371, opened 2023-02-23
[explorer][delegation] promote staking feature visibility

jillxuu:
https://user-images.githubusercontent.com/121921928/221021998-f59032e4-199a-473d-adea-c0db358dd06e.mov


0xZihan (approved):
please make the space between icon and number twice as bigger than now, or the same as the space between "14" and "APT". "14 APT" is a whole so the icon should at least farther than "14" to "APT".

Also please share this to Lex

PR #375, opened 2023-02-24
[ANS] show primary name

0xZihan:
With ANS primary name launching, we want to show primary names for the accounts instead of a random name when there're multiple names under one account.

* when you have a primary name, show the primary name
Screen Shot 2023-02-24 at 3 38 35 PM
* when you don't have a primary name, just show one of the names you own
Screen Shot 2023-02-24 at 3 39 04 PM
* when you own multiple names, you can search with any of the name, but the primary name will be attached on the side of address in the search result
Screen Shot 2023-02-24 at 3 38 56 PM

### DIFF
--- a/src/api/hooks/useGetANS.ts
+++ b/src/api/hooks/useGetANS.ts
@@ -1,6 +1,7 @@
 import {useEffect, useState} from "react";
 import {NetworkName} from "../../constants";
 import {useGlobalState} from "../../GlobalState";
+import {fetchJsonResponse} from "../../utils";
 
 function getFetchNameUrl(
   network: NetworkName,
@@ -24,16 +25,14 @@ export function useGetNameFromAddress(address: string) {
     const primaryNameUrl = getFetchNameUrl(state.network_name, address, true);
     if (primaryNameUrl !== undefined) {
       const fetchData = async () => {
-        const primaryNameResponse = await fetch(primaryNameUrl);
-        const {name: primaryName} = await primaryNameResponse.json();
+        const {name: primaryName} = await fetchJsonResponse(primaryNameUrl);
 
         if (primaryName) {
           setName(primaryName);
         } else {
           const nameUrl =
             getFetchNameUrl(state.network_name, address, false) ?? "";
-          const nameResponse = await fetch(nameUrl);
-          const {name} = await nameResponse.json();
+          const {name} = await fetchJsonResponse(nameUrl);
           setName(name);
         }
       };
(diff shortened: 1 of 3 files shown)


src/api/hooks/useGetANS.ts
```diff
@@ -2,25 +2,40 @@ import {useEffect, useState} from "react";
...
   const [name, setName] = useState<string | undefined>();
-  const url = getFetchNameUrl(state.network_name, address);
 
   useEffect(() => {
-    if (url !== undefined) {
+    const primaryNameUrl = getFetchNameUrl(state.network_name, address, true);
+    if (primaryNameUrl !== undefined) {
       const fetchData = async () => {
-        const response = await fetch(url);
-        const {name} = await response.json();
```
jillxuu, line 22: saw there are three places(here, line 75-76, line 27-28) doing the same thing: fetch url and then get json str from url response. would suggest we consolidate all three of them by having a helper function perform this logic.

src/api/hooks/useGetANS.ts
```diff
@@ -2,25 +2,40 @@ import {useEffect, useState} from "react";
...
-        setName(name);
+        const primaryNameResponse = await fetch(primaryNameUrl);
+        const {name: primaryName} = await primaryNameResponse.json();
+
+        if (primaryName) {
+          setName(primaryName);
```
jillxuu, line 31: as discussed offline, primaryName has `.apt`, which we will want to remove when displaying the name on explorer.

src/api/hooks/useGetANS.ts
```diff
@@ -43,28 +58,25 @@ function getFetchAddressUrl(network: NetworkName, name: string) {
...
   network: NetworkName,
-): Promise<string | undefined> {
+): Promise<{address: string | undefined; primaryName: string | undefined}> {
   const searchableName = name.endsWith(".apt") ? name.slice(0, -4) : name;
   const addressUrl = getFetchAddressUrl(network, searchableName);
 
+  const notFoundResult = {address: undefined, primaryName: undefined};
   if (addressUrl === undefined) {
-    return undefined;
+    return notFoundResult;
```
jillxuu, line 67: 👍

jillxuu (approved):

jillxuu:
lets address the comments before merging!

src/api/hooks/useGetANS.ts
0xZihan: will change it in the ans repo

PR #378, opened 2023-03-03
[explorer] fix modules sidebar header

0xbe1:
before
image
after
image


src/pages/Account/Tabs/ModulesTab.tsx
```diff
@@ -126,14 +126,7 @@ function ModulesContentReworked({address}: {address: string}): JSX.Element {
```
jillxuu, line 126: this change lgtm. but the mobile view's broken, can we quickly fix it as well?

Screenshot 2023-03-03 at 9 58 27 AM

0xbe1: what's the expected look?

jillxuu (approved):

PR #383, opened 2023-03-07
[AB Testing] add demo experiment

0xZihan:
Set up A/B testing with StatSig and add a demo experiment.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -10,6 +10,7 @@ import {MartianWallet} from "@martianwallet/aptos-wallet-adapter";
 import {RiseWallet} from "@rise-wallet/wallet-adapter";
 import {SpikaWallet} from "@spika/aptos-plugin";
 import {FewchaWallet} from "fewcha-plugin-wallet-adapter";
+import {StatsigProvider} from "statsig-react";
 
 import * as Sentry from "@sentry/react";
 import {BrowserTracing} from "@sentry/tracing";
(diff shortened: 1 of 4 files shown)


src/pages/Account/Components/AccountAllTransactions.tsx
```diff
@@ -26,6 +27,12 @@ function RenderPagination({
   ) => {
     searchParams.set("page", newPageNum.toString());
     setSearchParams(searchParams);
+
+    // logging
+    Statsig.logEvent("go_to_new_page", newPageNum, {
```
jillxuu, line 32: do we wanna have a separate constants file to keep track of all log events for statssig?

jillxuu: also will we have duplicated logs all over the place if we don't put this piece of logics in `useEffect`? with every re-render, this event will be logged?

src/index.tsx
```diff
@@ -59,13 +60,22 @@ const wallets = [
...
-        </BrowserRouter>
-      </AptosWalletAdapterProvider>
-    </QueryClientProvider>
+    <StatsigProvider
+      sdkKey={process.env.REACT_APP_STATSIG_SDK_KEY || ""}
+      waitForInitialization={true}
+      options={{
+        environment: {tier: "production"},
+      }}
+      user={{}}
```
jillxuu, line 69: could add in the future once we launch wallet connection in mainnet but we prob wanna use wallet address, etc as the user identification here?

jillxuu (commented):
statsig looks slick!

src/index.tsx
0xZihan: yes for features with wallet connection we'll need to include user. but for most of explorer features like this demo we don't need it. let's add it upon needed.

src/pages/Account/Components/AccountAllTransactions.tsx
0xZihan: yes it would be best to have a constants file to track them. this PR is for trying statsig out so we can maybe think more about this later.

0xZihan: It won't have duplicated logs because the log will only be triggered when user clicks on a page number and goes to a new page, which is the expected behaviour.

jillxuu (approved):
ah just saw its a demo experiment! lgtm

PR #387, opened 2023-03-09
[explorer] write contract

0xbe1:
image

In follow-up PRs, I will:
- Divide code in `ModulesTab` into a few smaller components
- Maybe prompt a toast on transaction submitted? Need design, and may need to change `useSubmitTransaction`
- Verify edge cases work, for example, `code::publish_package`

### DIFF
--- a/src/pages/Account/Tabs/ModulesTab.tsx
+++ b/src/pages/Account/Tabs/ModulesTab.tsx
@@ -1,5 +1,13 @@
 import {Types} from "aptos";
-import {Box, Grid, Stack, Typography, useTheme} from "@mui/material";
+import {
+  Box,
+  Button,
+  Grid,
+  Stack,
+  TextField,
+  Typography,
+  useTheme,
+} from "@mui/material";
 import React, {useState} from "react";
 import {orderBy} from "lodash";
 import SyntaxHighlighter from "react-syntax-highlighter";
@@ -22,6 +30,11 @@ import StyledTabs from "../../../components/StyledTabs";
 import StyledTab from "../../../components/StyledTab";
 import {useGetAccountModules} from "../../../api/hooks/useGetAccountModules";
 import {useGetAccountModule} from "../../../api/hooks/useGetAccountModule";
+import {WalletConnector} from "@aptos-labs/wallet-adapter-mui-design";
+import {useGlobalState} from "../../../GlobalState";
+import {useWallet} from "@aptos-labs/wallet-adapter-react";
+import {Controller, SubmitHandler, useForm} from "react-hook-form";
+import useSubmitTransaction from "../../../api/hooks/useSubmitTransaction";
 
 type PackageMetadata = {
   name: string;
@@ -29,6 +42,11 @@ type PackageMetadata = {
     name: string;
     source: string;
   }[];
+};
+
+type WriteContractFormType = {
+  typeArgs: string[];
+  args: string[];
 };
 
 const CONTRACT_ACTIONS = ["View code", "Write"] as const;
@@ -50,6 +68,13 @@ interface ModuleContentProps {
   address: string;
   moduleName: string;
   sourceCode: string;
+}
+
+interface WriteContractSidebarProps {
+  selectedModuleName: string | undefined;
+  selectedFnName: string | undefined;
+  moduleAndFnsGroup: Record<string, Types.MoveFunction[]>;
+  handleClick: (moduleName: string, fnName: string) => void;
 }
 
 function ModulesContent({address}: {address: string}) {
@@ -109,6 +134,7 @@ function ModulesReworked({address}: {address: string}): JSX.Element {
         setAction={setAction}
       />
       {action === "View code" && <ViewCode address={address} />}
+      {action === "Write" && <WriteContract address={address} />}
     </Box>
   );
 }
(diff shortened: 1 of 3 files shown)


src/pages/Account/Tabs/ModulesTab.tsx
```diff
@@ -179,6 +198,230 @@ function ViewCode({address}: {address: string}): JSX.Element {
...
+  if (error) {
+    return <Error address={address} error={error} />;
+  }
+
+  const modules = data ?? [];
+  if (modules.length === 0) {
+    return <EmptyTabContent />;
+  }
+
+  const moduleAndFnsGroup = modules.reduce((acc, module) => {
```
jillxuu, line 219: 👍👍

src/pages/Account/Tabs/ModulesTab.tsx
```diff
@@ -179,6 +198,230 @@ function ViewCode({address}: {address: string}): JSX.Element {
...
+          selectedFnName={selectedFnName}
+          moduleAndFnsGroup={moduleAndFnsGroup}
+          handleClick={(moduleName: string, fnName: string) => {
+            setSelectedModuleName(moduleName);
+            setSelectedFnName(fnName);
+          }}
+        />
+      </Grid>
+      <Grid item md={9}>
+        {!module || !fn ? (
```
jillxuu, line 257:
```suggestion
        {!module ? (
```
no need to check for `fn` anymore? as long as `module` doesn't exist, `fn`'s def not selected; vice versa, as long as `module` is selected, we can display the form

src/pages/Account/Tabs/ModulesTab.tsx
```diff
@@ -179,6 +198,230 @@ function ViewCode({address}: {address: string}): JSX.Element {
...
+function WriteContractSidebar({
+  selectedModuleName,
+  selectedFnName,
+  moduleAndFnsGroup,
+  handleClick,
+}: {
+  selectedModuleName: string | undefined;
+  selectedFnName: string | undefined;
+  moduleAndFnsGroup: Record<string, Types.MoveFunction[]>;
+  handleClick: (moduleName: string, fnName: string) => void;
```
jillxuu, line 276: nit: make it a function prop instead?

src/pages/Account/Tabs/ModulesTab.tsx
```diff
@@ -179,6 +198,230 @@ function ViewCode({address}: {address: string}): JSX.Element {
...
+                    label={`arg${hasSigner ? i - 1 : i}: ${param}`}
+                    fullWidth
+                  />
+                )}
+              />
+            );
+          })}
+        </Stack>
+        {connected ? (
+          <Button type="submit" variant="contained">
```
jillxuu, line 409: Screenshot 2023-03-09 at 3 39 20 PM
lets fix the button width?

src/pages/Account/Tabs/ModulesTab.tsx
```diff
@@ -1,5 +1,13 @@
 import {Types} from "aptos";
-import {Box, Grid, Stack, Typography, useTheme} from "@mui/material";
```
jillxuu, line 2: https://user-images.githubusercontent.com/121921928/224185381-a33ce61f-b805-4861-80ae-a9066c9b7a49.mov

LGTM! great manipulation on modules and functions. 

one general nit is that when i refresh the page, it will redirect me back to the initial stage where "view code" selected. would love to keep stay on the same tab and same function's selected as well.

src/pages/Account/Tabs/ModulesTab.tsx
```diff
@@ -179,6 +198,230 @@ function ViewCode({address}: {address: string}): JSX.Element {
...
+                <TextField
+                  onChange={onChange}
+                  value={value}
+                  label={`T${i}`}
+                  fullWidth
+                />
+              )}
+            />
+          ))}
+          {fn.params.map((param, i) => {
```
jillxuu, line 379: https://user-images.githubusercontent.com/121921928/224185683-c04cd52f-be1e-4eb0-8719-a316179f8fff.mov

seems like i can't perform the write?

jillxuu: this actually prompts me to think: for any kinds of contract, "write" might not be that helpful if we don't have proper "read". in this delegation example, if i wanna `unlock`, i need to know how much i have in there for me to unlock right? otherwise, it's gonna be tab switching, searching and digging to figure out the number first, and then perform the write later. couldn't figure out an ideal way to handle this, cuz there's no very systematic way to connect btw "unlock" and "active funds you have". 
But one thing that might be helpful here is: when tab switching, we should still keep the value in the input. so that if someone just wants to go to "Read" to checkout their amount, they can do so without repetitively entering their address again and again. @0xbe1 

cc @0xZihan

jillxuu (changes requested):

src/pages/Account/Tabs/ModulesTab.tsx
0xbe1: logically yes, however if we only check `!module ?` then `fn` has type `MoveFunction | undefined` which makes the code a bit difficult to read

the root cause is, we have not encoded the knowledge that "when fn is not null, then module must be not null as well" into TypeScript types.

We can do that though, by locating module and fn into just 1 type such as

```
type ModuleFn = {
  module: MoveModule;
  fn: MoveFunction;
}
```

and we do

```
const [moduleFn, setModuleFn] = useState<ModuleFn>();
```

wdyt? @jillxuu

src/pages/Account/Tabs/ModulesTab.tsx
0xbe1: that's where URL helps.

For example:

```
/account/[account]/modules/code/?module=[module]
```

and

```
/account/[account]/modules/write/?module=[module]&fn=[fn]
```

otherwise need to store info in local storage, which I think less optimal.

however, this routing change would be a larger scope change...

thoughts? @jillxuu @0xZihan

jillxuu: URL makes sense to me.

src/pages/Account/Tabs/ModulesTab.tsx
jillxuu: hmm seems like a overkill. i'm ok with the current implementation

src/pages/Account/Tabs/ModulesTab.tsx
0xbe1: believe this kind of routing change is out of scope - can we worry about it later?

src/pages/Account/Tabs/ModulesTab.tsx
0xbe1: prefer storing the info in URL - can we worry about this routing change later? @jillxuu

src/pages/Account/Tabs/ModulesTab.tsx
jillxuu: add url param strings is out of scope, but adding tabs in the param should be in scope so that when tab switching, url will change. 

https://user-images.githubusercontent.com/121921928/224802942-da90ac29-66a5-49a8-95fd-6568f872eb54.mov

jillxuu (approved):
overall lgtm. left one more comment. please address it before merging!

src/pages/Account/Tabs/ModulesTab.tsx
jillxuu: you could add something like `<Route path=":address/:tab/:action" element={<AccountPage />} />`  to the `/account` routing, and read the action from the param like `const {action} = useParams();`

0xbe1: @jillxuu this maybe find for now, but `:address/:tab/:action` seems too general 🤔 it seems more appropriate to add `:/address/modules/:tab` routes? wdyt

jillxuu: yeah this makes sense cuz it's stricter routing wise. feel free to go with this approach

0xbe1:
> overall lgtm. left one more comment. please address it before merging!

synced, decide to merge in first

PR #397, opened 2023-03-14
[explorer] Set default color scheme to system if not specified

moonclavedev:
### Color mode should be decentralized. 

This fix sets the default color scheme to the user's system settings if not previously toggled. 

You can test in Dev Tools by clearing all LocalStorage and making a selection under `Emulate CSS media feature prefers-color-scheme` within the Rendering tab. 

image

### DIFF
--- a/src/context/color-mode/ProvideColorMode.State.ts
+++ b/src/context/color-mode/ProvideColorMode.State.ts
@@ -7,34 +7,43 @@ export interface ColorModeContext {
   toggleColorMode: () => void;
 }
 
-type Mode = "light" | "dark";
+type Mode = "light" | "dark" | "system";
 
 const useProvideColorMode = () => {
-  const prefersDarkMode = useMediaQuery("(prefers-color-scheme: dark)", {
-    noSsr: true,
-  })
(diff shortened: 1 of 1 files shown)


0xZihan (approved):

src/context/color-mode/ProvideColorMode.State.ts
```diff
@@ -7,34 +7,43 @@ export interface ColorModeContext {
...
     setMode((prevMode) => {
       if (prevMode === "light") {
         localStorage.setItem("color_scheme", "dark");
         return "dark";
-      } else {
+      } else if (prevMode === "dark") {
         localStorage.setItem("color_scheme", "light");
         return "light";
+      } else {
+        localStorage.removeItem("color_scheme");
```
geekflyer, line 33: I think there is a bug here, which I also was able to confirm via the preview.
In the initial state, that is the first time a user presses the toggle it will go into the `else` branch of the code and set the "new theme" to the current system theme (aka nothing happens). Only when pressing the light/dark toggle a second time (and during the same browser session) it will actually toggle.

PR #399, opened 2023-03-15
[WIP] Introduce Code-Splitting, LazyLoad components, Lighthouse Improvements

moonclavedev:
### Code Splitting

- Integrate basic [code splitting](https://reactjs.org/docs/code-splitting.html) using [React.Lazy](https://reactjs.org/docs/code-splitting.html#reactlazy) dynamic imports mostly on a route-centric approach for testing. This results in a reduction of `main.js` by `-749.01 kB`. Ideally, the dynamic imports should be done on a container component basis (with skeleton loaders as placeholder), rather than the entire route. 

Breaking a large JavaScript file into smaller chunks results in a reduction of initial load time, improved caching, parallelism can be leveraged with HTTP/2, and critical files can be prioritized.

Before:
image

After:
image

Before:
image

After:
image

### Dependencies / 3rd Party JS

- Switch from [Lodash](https://www.npmjs.com/package/lodash) to [Lodash-es](https://www.npmjs.com/package/lodash-es), which is a modularized version of the library optimized for modern ES6 modules and tree-shaking capabilities. Regular Lodash is not optimized for tree-shaking and may include unnecessary code in the final bundle (saves about 30kb).
- Incorporate [react-helmet-async](https://www.npmjs.com/package/react-helmet-async) to manage 3rd-party scripts, eliminate render-blocking resources, and provide a simple method for page metadata on all routes

- TODO: Clean things up by moving individual inline scripts into their own components

### Replace ReactDOM.render with createRoot()

- ReactDOM.render is no longer supported in React 18 and is considered legacy, switch to [createRoot()](https://beta.reactjs.org/reference/react-dom/client/createRoot)

### Remove unused fonts loaded from Adobe Typekit

- We had about 6 unused font weights significantly adding to overall page size

### Additional Lighthouse Tweaks

- Table Headings are no longer `h6` elements, as this results in a score drop as headings being used out of order. Using heading tags in the correct order improves the user experience for screen readers and helps search engines better understand page content.

- Incorporate `aria-label` for links to improve accessibility score

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,8 +19,8 @@
     "@spika/aptos-plugin": "^0.1.2",
     "@trustwallet/aptos-wallet-adapter": "^0.1.6",
     "@types/lodash": "^4.14.191",
-    "@types/react": "latest",
-    "@types/react-dom": "latest",
+    "@types/react": "^18.0.28",
+    "@types/react-dom": "^18.0.11",
     "@types/react-simple-maps": "^3.0.0",
     "aptos": "^1.7.1",
     "aptos-node-checker-client": "0.0.2",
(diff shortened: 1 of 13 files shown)


kent-white:
closing this as stale

PR #401, opened 2023-03-15
[delegation] update banner with new design

jillxuu:
### FUTURE WORK
1. add side panel and click on banner will open up side panel
2. add statsig
3. mobile design still pending, add it later 

Screenshot 2023-03-15 at 2 03 52 PM

Screenshot 2023-03-15 at 2 03 57 PM


src/components/Banner.tsx
```diff
@@ -0,0 +1,97 @@
...
+              sx={{
+                backgroundColor: "#8B5CF6",
+                color: "#ffffff",
+                borderRadius: 1,
+                paddingX: 1,
+              }}
+            >
+              NEW
+            </Typography>
+            <Typography color="#ffffff">{children}</Typography>
```
angieyth, line 91: what do we do with this children here?

src/components/Banner.tsx
```diff
@@ -0,0 +1,97 @@
...
+          sx={{backgroundImage: `url(${AptosBannerImage})`, borderRadius: 1}}
+          icon={false}
+          action={
+            <Stack
+              direction="row"
+              spacing={1}
+              marginRight={2}
+              sx={{verticalAlign: "center"}}
+            >
+              <Button variant="text" onClick={handleClick}>
```
angieyth, line 50: what does this button do. does that mean it shows the same thing when clicking the button and clicking the banner itself?

moonclavedev:
Looking good! 

Couple things:

- I believe the original design uses Apparat for the main text copy, can we fix this? 
- Is the mobile design intended to be added later after this is merged, or before?

angieyth (approved):
this is exactly what I need for the ANS dogfooding! love love love

jillxuu:
> Is the mobile design intended to be added later after this is merged, or before?

mobile's not ready design wise. will add it later. but i do tried to make it mobile viewable, also everything's gated in flag. so should be safe

src/components/Banner.tsx
jillxuu: this children is the one that's passed in the props. its like

```
function Banner()  {
  return 
   <Typography
      color="#ffffff"
      fontFamily={theme.fontFamily.typography}
    >
      {children}
   </Typography>
}
```

and then we do something like 
```
<Banner>Hello</Banner>
```

With the above case, `children` = "Hello"

src/components/Banner.tsx
jillxuu: right now it doesn't do anything, cuz we don't have the side panel created yet. but basically once side panel's added, when users click on this button, it will open up side panel with delegation FAQs.

PR #407, opened 2023-03-17
[delegation] add side panel

jillxuu:
https://user-images.githubusercontent.com/121921928/226000134-e34bbcd6-bb7f-41de-8d4d-3ab99aa4078d.mov

### DIFF
--- a/src/pages/Validators/StakingDrawer.tsx
+++ b/src/pages/Validators/StakingDrawer.tsx
@@ -1,0 +1,127 @@
+import {
+  Button,
+  Divider,
+  Link,
+  List,
+  ListItem,
+  Typography,
+  useTheme,
+} from "@mui/material";
+import * as React from "react";
+import SideDrawer from "../../components/SideDrawer";
+import {REWARDS_LEARN_MORE_LINK} from "./Components/Staking";
+
+type StakingDrawerProps = {
+  open: boolean;
+  handleClick: () => void;
+};
+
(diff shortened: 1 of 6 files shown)


src/pages/Validators/StakingDrawer.tsx
```diff
@@ -0,0 +1,127 @@
...
+        <Typography variant="h4" sx={{margin: 1, marginBottom: 3}}>
+          Delegated Staking FAQ
+        </Typography>
+        <Divider flexItem sx={{marginY: 1}} />
+        <ListItem
+          sx={{
+            flexDirection: "column",
+            alignItems: "flex-start",
+            padding: 1,
+            color: theme.palette.mode === "dark" ? "#D4D4D4" : null,
```
angieyth, line 34: In ANS I usually see we pull color from mui material like `colors.grey[500]` and the color is set in the color scheme as a central place? do you think we should do that here too?

src/pages/Validators/StakingDrawer.tsx
```diff
@@ -0,0 +1,127 @@
...
+          </Typography>
+        </ListItem>
+        <Divider flexItem sx={{marginY: 1}} />
+        <ListItem
+          sx={{
+            flexDirection: "column",
+            alignItems: "flex-start",
+            padding: 1,
+            color: theme.palette.mode === "dark" ? "#D4D4D4" : null,
+          }}
```
angieyth, line 72: should we make this a const since we reuse this several times?

PR #408, opened 2023-03-17
[delegation] add mobile view to banner

jillxuu:
https://user-images.githubusercontent.com/121921928/228364626-9c394724-134e-4038-9a08-745992c05311.mov


moonclavedev (changes requested):

moonclavedev (dismissed):
Almost there! 

Can we compare to what's in the Figma file for mobile design? Specifically:

- Margin to the left of the content
- Tertiary button visual alignment with the content (going to require negative translate-x or margin, I suggest something like `transform: translateX(-0.5rem);`
- Font weight appears to be bold, I believe it's normal weight in the Figma file

Thanks!

jillxuu:
> Almost there!
> 
> Can we compare to what's in the Figma file for mobile design? Specifically:
> 
> * Margin to the left of the content
> * Tertiary button visual alignment with the content (going to require negative translate-x or margin, I suggest something like `transform: translateX(-0.5rem);`
> * Font weight appears to be bold, I believe it's normal weight in the Figma file
> 
> Thanks!

thanks for the suggestion! super helpful! addressed first two feedbacks. for third one, i believe font weight is 400, which is normal

0xZihan (approved):
Accept to unblock. LGTM

PR #413, opened 2023-03-20
[explorer] turn module sidebar from list into dropdown on narrow screen

0xbe1:


src/pages/Account/Tabs/ModulesTab.tsx
```diff
@@ -453,21 +459,37 @@ function ModuleSidebar({
...
+              name={moduleName}
+            />
+          ))}
+        </Box>
+      ) : (
+        <FormControl fullWidth>
+          <Select
+            value={selectedModuleIndex}
+            onChange={(e) => setSelectedModuleIndex(Number(e.target.value))}
+            sx={{textTransform: "none"}}
```
0xbe1, line 483: this doesn't seem to work though ... see "after" screenshot @jillxuu any help?

jillxuu: its prob cuz in Select you set the index instead of the value itself, so that `textTransform` wasn't directly applying on the module name

jillxuu (approved):

PR #421, opened 2023-03-21
[explorer] refactor modules tabs to reuse existing pattern

0xbe1:
/account/0x1/modules/code
image

/account/0x1/modules/write
image

/account/0x1/modules
image

/account/0x1/resources (still working)
image


jillxuu:
you are running into this problem cuz you name the module tab as `:tab`, which's exactly the same param name as the account `:tab` in this URL `<Route path=":address/:tab" element={<AccountPage />} />`. to fix it, you just need to update the naming to something like `<Route path=":address/modules/:modulesTab" element={<AccountPage />} />` in routing

PR #423, opened 2023-03-21
[delegation] update testnet delegation address to 0x1

jillxuu:
1. use 0x1 since delegation's in released in testnet
2. first check whether validator has delegation pool before querying all the delegation related info. so that we don't run into tons of network errors that will lead to 429
3. use whitelisted validator address again so that we can avoid querying view function on whether delegation pool exists or not for all validators, to save network query.

Screenshot 2023-03-22 at 12 43 35 PM

### DIFF
--- a/src/constants.tsx
+++ b/src/constants.tsx
@@ -68,17 +68,4 @@ export const defaultFeature = features[defaultFeatureName];
 /**
  * Delegation Service
  */
-
-// TODO(jill): update this function name once in mainnet
-export const DELEGATION_POOL_ADDRESS =
-  process.env.REACT_APP_DELEGATION_POOL_ADDRESS ??
-  "0x1310dc820487f24755e6e06747f6582118597a48868e2a98260fa8c3ee945cbd";
 export const OCTA = 100000000;
-export const MIN_ADD_STAKE_AMOUNT = 1; // min stake amount is 1 APT
-export const WHILTELISTED_TESTNET_DELEGATION_NODES = process.env
-  .REACT_APP_WHILTELISTED_TESTNET_DELEGATION_NODES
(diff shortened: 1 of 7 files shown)


src/pages/Validators/ValidatorsTable.tsx
```diff
@@ -163,7 +163,7 @@ function ValidatorHeaderCell({
   }
 }
 
-export type ValidatorCellProps = {
+type ValidatorCellProps = {
```
angieyth, line 166: why do we have two `type ValidatorCellProps`? one in src/pages/Validators/DelegationValidatorsTable.tsx:L224?

src/constants.tsx
```diff
@@ -68,17 +68,4 @@ export const defaultFeature = features[defaultFeatureName];
 /**
  * Delegation Service
  */
-
-// TODO(jill): update this function name once in mainnet
-export const DELEGATION_POOL_ADDRESS =
-  process.env.REACT_APP_DELEGATION_POOL_ADDRESS ??
-  "0x1310dc820487f24755e6e06747f6582118597a48868e2a98260fa8c3ee945cbd";
 export const OCTA = 100000000;
-export const MIN_ADD_STAKE_AMOUNT = 1; // min stake amount is 1 APT
```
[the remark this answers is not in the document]
jillxuu: now minimum is 10 APT. i forgot to delete this const before so deleting it here now seems really out of context lol

src/pages/Validators/ValidatorsTable.tsx
jillxuu: cuz we have two tables. one is validator table for all nodes, another is delegation validator table. before, since they don't differentiate in props, i re-used them, but now need to separate them up.

0xZihan (approved):

PR #427, opened 2023-03-23
[delegation] add inactive validator scenario for countdown bar

jillxuu:
### this pr
- when lockup cycle only exists in the past, meaning that validator's no longer active and lockup cycle will not get updated anymore. this pr is to add this edge case in.

### future PR
- need to use view function to check whether funds are withdrawble. cuz once validator is in inactive status, funds will not be moved from pending_inactive to inactive pool automatically, since the move is initiated with epoch change and lockup cycle change but lockup cycle will not be changed for inactive validators. 

Screenshot 2023-03-23 at 3 57 47 PM

### DIFF
--- a/src/components/IntervalBar.tsx
+++ b/src/components/IntervalBar.tsx
@@ -36,7 +36,10 @@ export default function IntervalBar({
     seconds: number;
     completed: boolean;
   }) => {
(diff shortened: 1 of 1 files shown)


angieyth (approved):
lgtm

PR #428, opened 2023-03-28
[explorer] use the same background color for source code and json view

0xbe1:
image
image
image
image


0xZihan (commented):
You can try:
1. Add a box in the modal as background.
2. Use some color picker to identify the RGB color of the code background in both dark and light mode, and use these colors instead of the RGBA colors.

src/pages/Account/Tabs/ModulesTab.tsx
```diff
@@ -662,7 +662,7 @@ function Code({bytecode}: {bytecode: string}) {
             style={
               theme.palette.mode === "light" ? solarizedLight : solarizedDark
             }
-            customStyle={{margin: 0}}
+            customStyle={{margin: 0, backgroundColor: "rgba(14,165,233,0.1)"}}
```
0xZihan, line 665: can we reuse the color constant?

jillxuu (approved):

PR #433, opened 2023-03-29
add tooltip to countdown interval when it is completed

jillxuu:
### future work
- once we have a list of inactive validators, will signal in the validator page that this validator is inactive and avoid showing the tooltip.

Screenshot 2023-03-29 at 11 00 22 AM


angieyth (approved):
code is good but I wonder if tooltip is the best format for this? it looks a little bit weird to show a tooltip there.

jillxuu:
> code is good but I wonder if tooltip is the best format for this? it looks a little bit weird to show a tooltip there.

will work with Lex on the format. any good suggestions here?

PR #441, opened 2023-04-04
[GP-12] support ANS search with partial APT suffix

0xaptosj:
Let's say an user's ANS name is `name.apt`, previous we can only search it by `name` or `name.apt`, this PR let user search by `name`, `name.`, `name.a`, `name.ap`, `name.apt`

### DIFF
--- a/src/api/hooks/useGetANS.ts
+++ b/src/api/hooks/useGetANS.ts
@@ -1,7 +1,7 @@
 import {useEffect, useState} from "react";
 import {NetworkName} from "../../constants";
 import {useGlobalState} from "../../GlobalState";
-import {fetchJsonResponse} from "../../utils";
+import {fetchJsonResponse, truncateAptSuffix} from "../../utils";
 
 function getFetchNameUrl(
   network: NetworkName,
@@ -52,13 +52,6 @@ function getFetchAddressUrl(network: NetworkName, name: string) {
   }
 
   return `https://www.aptosnames.com/api/${network}/v1/address/${name}`;
-}
-
-function truncateAptSuffix(name: string): string {
-  return name.replace(
-    /^[a-z\d][a-z\d-]{1,61}[a-z\d](\.apt|\.ap|\.a|\.?)$/,
-    "$1",
-  );
 }
 
 export async function getAddressFromName(
--- a/src/api/hooks/useGetANS.ts
+++ b/src/api/hooks/useGetANS.ts
@@ -54,11 +54,18 @@ function getFetchAddressUrl(network: NetworkName, name: string) {
   return `https://www.aptosnames.com/api/${network}/v1/address/${name}`;
 }
 
+function truncateAptSuffix(name: string): string {
+  return name.replace(
+    /^[a-z\d][a-z\d-]{1,61}[a-z\d](\.apt|\.ap|\.a|\.?)$/,
+    "$1",
+  );
+}
+
 export async function getAddressFromName(
   name: string,
   network: NetworkName,
 ): Promise<{address: string | undefined; primaryName: string | undefined}> {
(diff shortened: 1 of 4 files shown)


0xaptosj:
should i move the `yarn.lock` update to a separate PR?

src/api/hooks/useGetANS.ts
```diff
@@ -54,11 +54,21 @@ function getFetchAddressUrl(network: NetworkName, name: string) {
   return `https://www.aptosnames.com/api/${network}/v1/address/${name}`;
 }
 
+function truncateAptSuffix(name: string): string {
```
jillxuu, line 57: thanks for working on this issue! wondering can we use regex instead to find all the matching scenarios where "cathy.", "cathy.a", "cathy.ap", "cathy.apt" are all qualified and can be translated to "cathy"? 

some regex like `const regex = /^([a-zA-Z0-9-]+)(\.apt|\.ap|\.a|\.?)$/;`

0xaptosj: i feel like if we introduce more characters to ANS in the future, the regex will be more and more complex? generally i tried to avoid using regex because of it's difficult to read or test. thoughts?

jillxuu: in the forseeable future, we have primary names which is like "cathy.apt" and subdomains, which is like  "cathy.sub.apt". right now we only display primary names, and regex won't get complex as long as it fulfills the requirement "Name should have between 3 to 63 characters and contain only: lowercase (a-z), numbers (0-9), hyphen (-). A name may not start or end with a hyphen", which ANS uses regex to check against input string's validity. 

let me know if you have a strong preference here! we can discuss async as well

0xmaayan: This is the regex we widely use 

 /^[a-z\d][a-z\d-]{1,61}[a-z\d]$/;

I think it is somewhere is the ANS codebase, @angieyth might be nice to export it as a util function from the sdk - something like isNameValid

yarn.lock
```diff
@@ -38,13 +38,14 @@
     tslib "^2.3.0"
     zen-observable-ts "^1.2.5"
 
-"@aptos-labs/aptos-names-connector@^0.0.1":
-  version "0.0.1"
-  resolved "https://registry.yarnpkg.com/@aptos-labs/aptos-names-connector/-/aptos-names-connector-0.0.1.tgz#51f5ea6f6fc80b679ed4b03a29f35623231bcd25"
-  integrity sha512-HTVqbnsaKyfaqGdf7fC9dJsj8hr9aLS4mB5XUb8KxN+o/KKW4Ui0KpoPeNoLasCC+CZAcQohCvcPWJeRd5UNsg==
+"@aptos-labs/aptos-names-connector@^0.0.2":
```
0xmaayan, line 41: Why these changes?

0xmaayan (changes requested):
Highly recommend adding some unit tests to test these changes

yarn.lock
0xaptosj: last [PR](https://github.com/aptos-labs/explorer/pull/440) only updated `package.json` but not `yarn.lock`

0xaptosj:
> Highly recommend adding some unit tests to test these changes

do we have any ts repo with unit test i can refer to? im new to frontend

0xmaayan:
> > Highly recommend adding some unit tests to test these changes
> 
> 
> 
> do we have any ts repo with unit test i can refer to? im new to frontend 🙂️

Can take a look here https://github.com/aptos-labs/governance/blob/main/src/__tests__/utils.test.tsx

0xmaayan (approved):

PR #444, opened 2023-04-06
[code view] Implement deep linking to module and function name

geekflyer:
This PR implements deep linking from entry function links in the transaction list and transaction details to the corresponding module and function name.

### Motivation
Often times when looking at a list of transactions it's useful to do what this transaction was actually doing by looking at the source code.
Currently it takes 5+ mouse clicks (to get the corresponding module) + several keystrokes (to copy&paste the function name and find it in the source code) in rather unintuitive places to get from a function name to it's definition. This PR reduces this in the ideal case to essentially 1 single mouse click.

### Some implementation notes:
1. Introduced a new Link component that tries to use Copyable Links / hrefs vs programmatic navigation so that users can right-click a URL and copy it or open it in a new tab vs having to actually click a button.
2. The currently used `<SyntaxHighlighter>` component for the module code doesn't make it super easy to scroll to a specific section out of the box, so I implemented some pixel based scrolling logic on the wrapping `<Box>` component. It's a bit ugly but it works for now.
3. Implemented some heuristic that finds the entry function in the source code, but that heuristic can be improved as described in the code comment via some regex or so (probably needs some regex safe escaping).

demo video:

https://user-images.githubusercontent.com/1221897/230337005-6fa28d22-ed4f-45ba-a969-3c772edac962.mp4

### DIFF
--- a/src/components/Table/GeneralTableCell.tsx
+++ b/src/components/Table/GeneralTableCell.tsx
@@ -10,6 +10,7 @@ interface GeneralTableCellProps extends TableCellProps {
 export default function GeneralTableCell({
   children,
   sx,
+  ...rest
 }: GeneralTableCellProps) {
   return (
     <TableCell
@@ -20,6 +21,7 @@ export default function GeneralTableCell({
         },
         ...(Array.isArray(sx) ? sx : [sx]),
       ]}
+      {...rest}
     >
       {children}
     </TableCell>
--- a/src/components/Table/GeneralTableCell.tsx
+++ b/src/components/Table/GeneralTableCell.tsx
@@ -10,7 +10,7 @@ interface GeneralTableCellProps extends TableCellProps {
 export default function GeneralTableCell({
   children,
   sx,
-  ...rest
+  ...props
 }: GeneralTableCellProps) {
   return (
     <TableCell
@@ -21,7 +21,7 @@ export default function GeneralTableCell({
         },
         ...(Array.isArray(sx) ? sx : [sx]),
       ]}
-      {...rest}
+      {...props}
     >
       {children}
     </TableCell>
--- a/src/components/router-link.tsx
+++ b/src/components/router-link.tsx
@@ -7,6 +7,7 @@ export function Link({to, children}: {to: string; children: React.ReactNode}) {
     <MuiLink
       component={RouterLink}
       to={to}
+      style={{textDecoration: "none", color: "inherit"}}
       onClick={(e) => {
         e.stopPropagation();
       }}
--- a/src/components/router-link.tsx
+++ b/src/components/router-link.tsx
@@ -1,0 +1,17 @@
+import {Link as MuiLink} from "@mui/material";
+import React from "react";
+import {Link as RouterLink} from "react-router-dom";
+
+export function Link({to, children}: {to: string; children: React.ReactNode}) {
+  return (
+    <MuiLink
+      component={RouterLink}
+      to={to}
+      onClick={(e) => {
+        e.stopPropagation();
+      }}
+    >
+      {children}
+    </MuiLink>
+  );
+}
--- a/src/pages/Account/Tabs/ModulesTab/Tabs.tsx
+++ b/src/pages/Account/Tabs/ModulesTab/Tabs.tsx
@@ -39,13 +39,13 @@ function TabPanel({value, address}: TabPanelProps): JSX.Element {
 function ModulesTabs({address}: {address: string}): JSX.Element {
   const theme = useTheme();
   const tabValues = Object.keys(TabComponents) as TabValue[];
-  const {modulesTab} = useParams();
+  const {selectedModuleName, modulesTab} = useParams();
   const navigate = useNavigate();
   const value =
     modulesTab === undefined ? tabValues[0] : (modulesTab as TabValue);
 
   const handleChange = (event: React.SyntheticEvent, newValue: TabValue) => {
-    navigate(`/account/${address}/modules/${newValue}`);
+    navigate(`/account/${address}/modules/${selectedModuleName}/${newValue}`);
   };
 
   return (
--- a/src/pages/Account/Tabs/ModulesTab/ViewCode.tsx
+++ b/src/pages/Account/Tabs/ModulesTab/ViewCode.tsx
@@ -270,6 +270,11 @@ function ModuleHeader({
   );
 }
 
+// inspired by https://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex
+function escapeRegExp(regexpString: string) {
+  return regexpString.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+}
+
 function useStartingLineNumber(sourceCode?: string) {
   const [searchParams] = useSearchParams();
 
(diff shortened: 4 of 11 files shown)


src/pages/Transaction/Tabs/Components/TransactionFunction.tsx
```diff
@@ -4,8 +4,7 @@ import {Types} from "aptos";
 import CurrencyExchangeOutlinedIcon from "@mui/icons-material/CurrencyExchangeOutlined";
 import DescriptionOutlinedIcon from "@mui/icons-material/DescriptionOutlined";
 import {CodeLineBox} from "../../../../components/CodeLineBox";
-
-const BLOCK_MODULE_NAME = "candy_machine_v2";
```
geekflyer, line 8: wasn't really sure what this BLOCK_MODULE_NAME special code for candy_machine_v2 is for and removed it. Please let me know if this needs to be added back and what this is for.

src/components/router-link.tsx
```diff
@@ -0,0 +1,17 @@
+import {Link as MuiLink} from "@mui/material";
+import React from "react";
+import {Link as RouterLink} from "react-router-dom";
+
+export function Link({to, children}: {to: string; children: React.ReactNode}) {
```
jillxuu, line 5: great idea on making these entry functions clickable. but i'm thinking instead of making the function name a deeplink, we instead could have a little icon next to it, and clicking the icon should lead us to the module page. reason why: we have this [design](https://www.figma.com/proto/gG4UP6OipyXq0LjM4zWmMw/Explorer-Design-Exploration?node-id=2-1602&starting-point-node-id=2%3A1602) for explorer improvement, where having a icon next to a field either to copy it or click it is gonna be more consistent.

src/pages/Account/Tabs/ModulesTab/Tabs.tsx
```diff
@@ -39,13 +39,13 @@ function TabPanel({value, address}: TabPanelProps): JSX.Element {
 function ModulesTabs({address}: {address: string}): JSX.Element {
```
jillxuu, line 39: switch btw "read" and "write" tab doesn't maintain the search param.

src/pages/Account/Tabs/ModulesTab/Tabs.tsx
```diff
@@ -39,13 +39,13 @@ function TabPanel({value, address}: TabPanelProps): JSX.Element {
...
   const tabValues = Object.keys(TabComponents) as TabValue[];
-  const {modulesTab} = useParams();
+  const {selectedModuleName, modulesTab} = useParams();
   const navigate = useNavigate();
   const value =
     modulesTab === undefined ? tabValues[0] : (modulesTab as TabValue);
 
   const handleChange = (event: React.SyntheticEvent, newValue: TabValue) => {
-    navigate(`/account/${address}/modules/${newValue}`);
+    navigate(`/account/${address}/modules/${selectedModuleName}/${newValue}`);
```
jillxuu, line 48: now clicking btw module names actually refresh the page, instead of just update the source code & abi on the right side. when we switch tabs in the account page, it does change url param but page's not refreshed. lets get it fixed!

src/pages/Account/Tabs/ModulesTab/ViewCode.tsx
```diff
@@ -67,44 +70,68 @@ function ViewCode({address}: {address: string}): JSX.Element {
...
+            moduleName={selectedModuleName}
+            bytecode={selectedModule.source}
+          />
+        )}
       </Grid>
     </Grid>
   );
 }
 
 function ModuleSidebar({
```
jillxuu, line 130: lets also sync with Lex on this. rn its not very intuitive that users can click on module name to open up another tab. having an icon next to it might be more helpful

src/pages/Account/Tabs/ModulesTab/ViewCode.tsx
```diff
@@ -244,6 +270,26 @@ function ModuleHeader({
...
+
+  if (!sourceCode) return 0;
+
+  const entryFunctionToHightlight = searchParams.get("entry_function");
+  if (entryFunctionToHightlight == null) return 0;
+
+  const lines = sourceCode.split("\n");
+  const lineNumber = lines.findIndex((line) =>
+    // TODO: improve line matching algorithm by using regex or something. Currently this will match falsely other entry functions that contain the entry function name in their name - e.g. `my_func_blue` will be matched when searching for `my_func`.
+    line.includes(`public entry fun ${entryFunctionToHightlight}`),
```
jillxuu, line 284: thats what gpt told me `^public\sentry\sfun\s\{\$\{entryFunctionToHighlight\}\}$` lol

jillxuu (commented):
thanks for the contribution! love this direction. 

on UI, can we still maintain the original UI details, like color, etc?

src/components/router-link.tsx
geekflyer: hmm, so you're saying clicking function name should do nothing?
I think it's still reasonable and better UX to have the function name itself clickable and perform a default action (i.e. in this case navigate to the function).
Also in the linked design I would suggest the same for the sender and sent to addresses - i.e. click on the address takes you to it, click on the copy icon next to it copies the full address.

src/pages/Account/Tabs/ModulesTab/Tabs.tsx
geekflyer: That's currently expected since search param has currently no function in the write tab (deep link support for the write tab is not implemented) and we currently don't link to the write tab from anywhere anyways.
It's certainly something we can consider to implement before graduating the "write" (or rather Run) functionality from dev mode to GA, but I was under the impression that the Run mode needs some further changes anyways so I didn't want to touch it too much (but also not break existing functionality).

src/pages/Account/Tabs/ModulesTab/Tabs.tsx
geekflyer: not sure I quite understand? I don't think it does a full page refresh - it visually looks all the same as before adding the deeplinking and it definately doesn't do a full page reload and reloads the JS assets.

If you can take a video of what you're seeing that might be helpful but if its easy to explain that's also fine.

jillxuu: lol i was playing around with something else in the dev env that mess with this `<Link>`. i can confirm it's all good, no page reload!

src/pages/Account/Tabs/ModulesTab/ViewCode.tsx
geekflyer: I think that's standard behaviour of browser links.
Take for example Github - I can right click in a lot of places to open something in a new tab (or window as desired):
image.png
It's a feature, not a bug 😄

src/components/Table/GeneralTableCell.tsx
```diff
@@ -10,6 +10,7 @@ interface GeneralTableCellProps extends TableCellProps {
 export default function GeneralTableCell({
   children,
   sx,
+  ...rest
```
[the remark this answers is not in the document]
geekflyer: done

src/components/Table/GeneralTableCell.tsx
```diff
@@ -20,6 +21,7 @@ export default function GeneralTableCell({
         },
         ...(Array.isArray(sx) ? sx : [sx]),
       ]}
+      {...rest}
```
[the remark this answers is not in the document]
geekflyer: done

src/pages/Account/Tabs/ModulesTab/ViewCode.tsx
geekflyer: hmm, that chatgpt output was pretty wrong in many ways, but I added something better :)

geekflyer:
> thanks for the contribution! love this direction.
> 
> on UI, can we still maintain the original UI details, like color, etc?

Addressed most of your feedback incl. retaining original design.
Side thing: I made the timestamps in the tx table display second precision which is what me and a few other folks requested in the past and I believe is appropriate for a blockchain that has sub-second latency.

geekflyer:
@jillxuu  implemented the suggested design changes to distinguish between clickable and not clickable. Also made coin transfer clickable.

jillxuu:
> @jillxuu implemented the suggested design changes to distinguish between clickable and not clickable. Also made coin transfer clickable.

lgtm! thanks for working on this!

jillxuu (approved):

src/pages/Transaction/Tabs/Components/TransactionFunction.tsx
0xZihan: Could you add it back? I'm following up in the slack.

PR #445, opened 2023-04-06
migrate from yarn to pnpm

geekflyer:
This PR switches the package manager from `yarn` to `pnpm` to align with aptos-core and the ops repo. Since pnpm doesn't flatten the dependency tree (unlike yarn) it turned out some undeclared dependencies that are currently used (like @mui/system and @jest/globals) were not explicitly declared in the package.json, so I added them.


angieyth:
should we always use pnpm instead of yarn for the packages in the future?

angieyth (approved):

geekflyer:
> should we always use pnpm instead of yarn for the packages in the future?

I don't want to be too prescriptive but I would recommend it. It's overall a bit faster, uses a lot less disk space and the commands are more similar to core `npm` which makes it a bit easier to switch between `npm` and `pnpm` whereas `yarn` CLI used a bit different naming for some commands and flags. We had a discussion about it here https://aptos-org.slack.com/archives/C03N83P7QUC/p1667514786627619 some time ago fyi.

PR #449, opened 2023-04-07
Fix network switching behaviour and remove selected network from localStorage

geekflyer:
Fix network switching behaviour and remove selected network from localStorage. Fixes GP-2 (check for reasoning).
Also makes sure that all navigation actions and links retain the selected network context when navigating within the site.
This should create more consistent/predictable experience when sharing links. Previously it could happen that someone shared a link that was taken from devnet/testnet/mainnet and once opened in another user's browser it would show up as empty.

before:

https://user-images.githubusercontent.com/1221897/231326140-6f964f3e-c3f0-4fef-a8bc-4ada48bcfab6.mp4

after:

https://user-images.githubusercontent.com/1221897/231326157-80cf696e-574d-4a39-9d3e-4b61bef3d4d9.mp4


0xZihan:
This is amazing! Thanks @geekflyer I've been wanting to do this for a while. @jillxuu this is quite a big change, let's maybe do some informal internal dogfooding with https://deploy-preview-449--aptos-explorer.netlify.app/ to test for a couple of days.

src/pages/Account/Components/TokensTable.tsx
```diff
@@ -16,11 +15,9 @@ type TokenCellProps = {
 function TokenNameCell({token}: TokenCellProps) {
   return (
     <GeneralTableCell sx={{textAlign: "left"}}>
-      <Link
-        component={RRD.Link}
+      <InternalLink
         to={`/token/${token?.token_data_id_hash}/${token?.property_version}`}
-        color="primary"
-        underline="none"
+        // color="primary"
```
jillxuu, line 20:
```suggestion
```

src/components/HashButton.tsx
```diff
@@ -37,17 +37,18 @@ function getHashLinkStr(hash: string, type: HashType): string {
...
+
   switch (type) {
     case HashType.ACCOUNT:
     case HashType.TRANSACTION:
       return (
-        <Link
-          component={RRD.Link}
-          to={getHashLinkStr(hash, type)}
+        <InternalLink
+          to={augmentToWithGlobalSearchParams(getHashLinkStr(hash, type))}
```
jillxuu, line 47: do we still need `augmentToWithGlobalSearchParams` here since we're already using `InternalLink`?

src/pages/Account/Tabs/ModulesTab/ViewCode.tsx
```diff
@@ -78,6 +82,7 @@ function ViewCode({address}: {address: string}): JSX.Element {
   );
 
   const navigate = useNavigate();
+  const augmentToWithGlobalSearchParams = useAugmentToWithGlobalSearchParams();
```
jillxuu, line 85: it's not used?

src/api/hooks/useGetSearchResults.ts
```diff
@@ -54,7 +57,7 @@ export default function useGetSearchResults(input: string) {
               label: `Account ${truncateAddress(address)}${
                 primaryName ? ` | ${primaryName}.apt` : ``
               }`,
-              to: `/account/${address}`,
+              to: augmentToWithGlobalSearchParams(`/account/${address}`),
```
jillxuu, line 60: couple other places needs to use `augmentToWithGlobalSearchParams` as well, i.e. https://github.com/aptos-labs/explorer/blob/91bd8d92df6bfd4e2261c6a5f17196555c4ca242/src/api/hooks/useGetSearchResults.ts#L90.

jillxuu (commented):
overall lgtm. left couple nit comments.

jillxuu:
> This is amazing! Thanks @geekflyer I've been wanting to do this for a while. @jillxuu this is quite a big change, let's maybe do some informal internal dogfooding with https://deploy-preview-449--aptos-explorer.netlify.app/ to test for a couple of days.

sg. as long as the logic works as expected, even if we miss out some places to keep searchparams in the url, that's fine, as we can always add them later once found.

src/components/HashButton.tsx
geekflyer: good find, nope

src/api/hooks/useGetSearchResults.ts
geekflyer: actually I'm already augmenting the url for _all_ results on line 152. That being said the augmentation on line 60 is kinda unnecessary since it's being double augmented, so I'm just removing it from line 60 to avoid confusion.

geekflyer:
solved by #464

PR #451, opened 2023-04-08
Fix ans name caching

0xaptosj:
We need to cache in `useEffect` otherwise we are caching the previous value because `useState` is not synchronous.

This will try to cache both sender and receiver if they have corresponding ANS, since we set cache expiration to 1 min, it shouldn't blow up the cache.


0xaptosj:
actually this will blowup the cache, because we do not remove the cache until it's hit again

jillxuu (approved):

PR #452, opened 2023-04-08
fix ANS caching

0xaptosj:
**Background**
We previously try to cache ANS name to reduce API call to indexer. But there were some issues.

**Solution**
We need to cache in `useEffect` otherwise we are caching the previous value because useState is not synchronous.

Since we only cache the account owner's ANS name, it shouldn't add too much data to `localStorage`, chatgpt said if average string length is 50 char, `localStorage` should be able to hold 100_000 strings.

### DIFF
--- a/src/components/TitleHashButton.tsx
+++ b/src/components/TitleHashButton.tsx
@@ -109,7 +109,7 @@ function HashButton({hash}: {hash: string}) {
 
 function Name({address}: {address: string}) {
   const theme = useTheme();
-  const name = useGetNameFromAddress(address);
+  const name = useGetNameFromAddress(address, true);
 
   if (!name) {
     return null;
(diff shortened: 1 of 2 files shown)


src/components/TitleHashButton.tsx
```diff
@@ -109,7 +109,7 @@ function HashButton({hash}: {hash: string}) {
 
 function Name({address}: {address: string}) {
   const theme = useTheme();
-  const name = useGetNameFromAddress(address);
+  const name = useGetNameFromAddress(address, true);
```
angieyth, line 112: why do we cache here but not in `HashButton.tsx` (https://github.com/aptos-labs/explorer/blob/36b777ef8fe80a3eee5cca0282b58ff6854a36e5/src/components/HashButton.tsx#L92)

angieyth (approved):
lgtm

src/components/TitleHashButton.tsx
0xaptosj: we try to get name of `HashButton.tsx` whenever there's an address, (e.g. on an account detail page, we try to get the name for sender / receiver of every tx), im concerned this will save too much stuff into the cache, plus i think only the account owner's name is the most commonly appeared one across the page, so i only cache the owner

PR #464, opened 2023-04-13
Fix network switching and link sharing by making sure network is always appended to the URL

geekflyer:
- Fix network switching behaviour and remove selected network from localStorage
- implement network switch with using localStorage setting as implicit user default

This PR is an evolution of #461 and retains some localStorage usage to maintain a user specific default network selection.


jillxuu (approved):
lgtm!

PR #465, opened 2023-04-13
[delegation] add validator state

jillxuu:
Screenshot 2023-04-13 at 1 07 07 PM
Screenshot 2023-04-13 at 1 05 36 PM

### DIFF
--- a/src/api/hooks/useGetDelegationNodeInfo.ts
+++ b/src/api/hooks/useGetDelegationNodeInfo.ts
@@ -1,6 +1,6 @@
 import {AptosClient, Types} from "aptos";
 import {useState, useMemo} from "react";
-import {getValidatorCommission} from "..";
+import {getValidatorCommission, getValidatorState} from "..";
 import {useGlobalState} from "../../GlobalState";
 import {useGetAccountResource} from "./useGetAccountResource";
 import {useGetValidatorSet} from "./useGetValidatorSet";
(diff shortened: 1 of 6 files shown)


src/api/hooks/useGetDelegationNodeInfo.ts
```diff
@@ -36,11 +37,13 @@ export function useGetDelegationNodeInfo({
   const [delegatedStakeAmount, setDelegatedStakeAmount] = useState<string>();
   const [networkPercentage, setNetworkPercentage] = useState<string>();
   const [isQueryLoading, setIsQueryLoading] = useState<boolean>(true);
+  const [validatorStatus, setValidatorStatus] = useState<Types.MoveValue[]>();
 
   useMemo(() => {
     if (!isLoading) {
       const fetchData = async () => {
         setCommission(await getValidatorCommission(client, validatorAddress));
+        setValidatorStatus(await getValidatorState(client, validatorAddress));
```
movekevin, line 46: Does this function handle pending active and pending inactive states as well?

movekevin (approved):

src/api/hooks/useGetDelegationNodeInfo.ts
jillxuu: i was hoping we don't overwhelm users too much with different validator states. if validator's inactive, we show them as inactive, the rest's shown as active. lmk if you don't think it's the best approach

PR #467, opened 2023-04-13
handle 429

jillxuu:
gracefully handle 429 with an error msg

https://aptos-org.slack.com/archives/C03EP0XM99D/p1681330225916179

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -1,11 +1,13 @@
 export enum ResponseErrorType {
   NOT_FOUND = "Not found",
   UNHANDLED = "Unhandled",
+  TOO_MANY_REQUESTS = "To Many Requests",
 }
 
 export type ResponseError =
   | {type: ResponseErrorType.NOT_FOUND; message?: string}
-  | {type: ResponseErrorType.UNHANDLED; message: string};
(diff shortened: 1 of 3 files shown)


src/api/client.ts
```diff
@@ -1,11 +1,13 @@
 export enum ResponseErrorType {
   NOT_FOUND = "Not found",
   UNHANDLED = "Unhandled",
+  TOO_MANY_REQUESTS = "To Many Requests",
```
0xaptosj, line 4: typo here

src/pages/Account/Error.tsx
```diff
@@ -8,22 +8,30 @@ type ErrorProps = {
...
+          Try again later
+        </Alert>
+      );
+    case ResponseErrorType.TOO_MANY_REQUESTS:
+      return (
+        <Alert severity="error">
+          To many requests. Please try again 5 minutes later.
+        </Alert>
+      );
   }
```
0xaptosj, line 36: do we need default case here? or it always falls under UNHANDLED if not 429 and not found

jillxuu: yeah UNHANDLED is the default experience rn. we don't easily use <Error> component only if thats needed, so UNHANDLED won't get triggered easily.

PR #468, opened 2023-04-14
pixelate account address

jillxuu:
Screenshot 2023-04-13 at 5 17 11 PM


geekflyer:
the copy address button doesn't work in the txn table. You probably have to add e.stopPropagation to the event handler to avoid it triggering the click handler of the overall row (this is similar to what I did for the entry function links).

geekflyer:
also btw I would hightly suggest to add an on-hover tooltip that shows the full address when you hover over it.

moonclavedev:
> the copy address button doesn't work in the txn table. You probably have to add e.stopPropagation to the event handler to avoid it triggering the click handler of the overall row (this is similar to what I did for the entry function links).

Correct!

geekflyer:
Lastly I think the account address should be clickable and navigate to the account.

jillxuu:
addressed feedbacks: https://deploy-preview-468--aptos-explorer.netlify.app/?network=mainnet

geekflyer (approved):
thanks for addressing the feedback!

Some minor nits:
1. I feel like the copy button currently is visually a bit too seperated from the address and takes up too much space. I think it might be better to make the copy button part of the address box instead of being a seperate thing next to it.
I.e. look at:
image
where the copy button is visually part of the address box.


2. Maybe add a small delay to the tooltip displaying. Based on some googling 300-500ms seems common.
image

Approving this now preemptively but would love if you could address it.

PR #473, opened 2023-04-19
enable wallet connection on mobile

jillxuu:
Screenshot 2023-04-19 at 12 40 37 PM
Simulator Screen Shot - iPhone 14 Pro Max - 2023-04-21 at 16 42 36

GP-26

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
   "dependencies": {
     "@apollo/client": "^3.7.9",
     "@aptos-labs/aptos-names-connector": "^0.0.2",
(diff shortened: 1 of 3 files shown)


0xmaayan:
fyi, currently the adapter shows all wallets on mobile as it doesn't know which one has mobile support and/or what in-app browser you use. So bottom line, you probably would end up showing all wallets in any in-app browser (most of them will be disabled though)

This is part of having mobile support which is in the roadmap and agree on having it ready by mid May. cc @gregnazario

kent-white (approved):

kent-white:
thank you!

PR #478, opened 2023-04-20
[delegation] fix validator state mismatch issue

jillxuu:
previously validator state's queried inside a different component, might cause mismatch when table's reordered.


movekevin (approved):

movekevin (commented):
Ideas for improvements later:
1. Add a tooltip if the validator is pending_active to tell users when they'll become active.
2. Similarly for pending inactive
3. Add a tooltip indicating how long a validator has stayed active
4. Add a tooltip showing the last time an inactive validator was active

PR #479, opened 2023-04-20
[delegation] fix table sorting

jillxuu:
https://deploy-preview-479--aptos-explorer.netlify.app/validators?network=mainnet

### DIFF
--- a/src/pages/Validators/DelegationValidatorsTable.tsx
+++ b/src/pages/Validators/DelegationValidatorsTable.tsx
@@ -56,11 +56,13 @@ function getValidatorsOrderedBy(
   column: Column,
 ) {
   switch (column) {
-    case "delegatedAmount":
-      return validatorsCopy.sort(
-        (validator1, validator2) =>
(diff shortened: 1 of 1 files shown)


src/pages/Validators/DelegationValidatorsTable.tsx
```diff
@@ -56,11 +56,13 @@ function getValidatorsOrderedBy(
   column: Column,
 ) {
   switch (column) {
-    case "delegatedAmount":
```
movekevin, line 59: Why not keep this as well? :) Sorting by total delegation amount is useful

jillxuu: `delegatedAmount` and other fields, like commission, delegator number, are either from view function or indexer query. i'm working on making all fields sortable but right now only column from json file's sortable, thats why only rewards earned can be sorted.

movekevin: I see

movekevin (approved):

PR #481, opened 2023-04-21
[delegation] fix address leading zero issue

jillxuu:
https://deploy-preview-481--aptos-explorer.netlify.app/validators/delegation?network=mainnet

reference: https://github.com/aptos-labs/internal-ops/pull/1239

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -107,3 +107,27 @@ export function truncateAptSuffix(name: string): string {
     "$1",
   );
 }
+
+/**
+ * Standardizes an address to the format "0x" followed by 64 lowercase hexadecimal digits.
+ */
+export const standardizeAddress = (address: string): string => {
+  // Convert the address to lowercase
+  address = address.toLowerCase();
+  // If the address has more than 66 characters, it's already invalid
+  if (address.length > 66) {
+    return address;
+  }
+  // Remove the "0x" prefix if present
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -107,3 +107,27 @@ export function truncateAptSuffix(name: string): string {
     "$1",
   );
 }
+
+/**
+ * Standardizes an address to the format "0x" followed by 64 lowercase hexadecimal digits.
+ */
+export const standardizeAddress = (address: string): string => {
```
movekevin, line 114: Is this consistent with the standardize address in the indexer? 

/// Standardizes all addresses and table handles to be length 66 (0x-64 length hash)
pub fn standardize_address(handle: &str) -> String {
    format!("0x{:0>64}", &handle[2..])
}

jillxuu: yeah i believe so. can i get additional eyes on this @bowenyang007 @larry-aptos  to make sure it aligns with indexer way of standardizing address? [example](https://github.com/aptos-labs/aptos-core/blob/e149dc164f1ea77d410ac8fc29780b17cc7ec778/crates/indexer/src/util.rs#L15)

larry-aptos: looks good. the code jill referring to takes invalid cases into consideration.

larry-aptos (approved):

PR #483, opened 2023-04-24
[explorer] /modules default to /modules/code

0xbe1:
as we're about to add read and write, the current routes for /modules doesn't work.

current:
/account/:account/modules/:module

proposed:
/account/:account/modules/:action/:module (action is one of "code", "read" and "write")

see https://deploy-preview-483--aptos-explorer.netlify.app/account/0x1/modules/code/account?network=mainnet for an example

### DIFF
--- a/src/ExplorerRoutes.tsx
+++ b/src/ExplorerRoutes.tsx
@@ -31,11 +31,7 @@ export default function ExplorerRoutes() {
         </Route>
         <Route path="/account">
           <Route
-            path=":address/modules/:selectedModuleName/:modulesTab"
-            element={<AccountPage />}
-          />
-          <Route
-            path=":address/modules/:selectedModuleName"
+            path=":address/modules/:modulesTab/:selectedModuleName"
             element={<AccountPage />}
           />
           <Route path=":address/:tab" element={<AccountPage />} />
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -169,6 +169,14 @@ export function getAccountModule(
   );
 }
 
+export function view(
+  request: Types.ViewRequest,
+  nodeUrl: string,
+): Promise<Types.MoveValue[]> {
+  const client = new AptosClient(nodeUrl);
+  return withResponseError(client.view(request));
+}
+
 export function getTableItem(
   requestParameters: {tableHandle: string; data: Types.TableItemRequest},
   nodeUrl: string,
--- a/src/pages/Account/Tabs.tsx
+++ b/src/pages/Account/Tabs.tsx
@@ -93,10 +93,10 @@ export default function AccountTabs({
   accountData,
   tabValues = TAB_VALUES,
 }: AccountTabsProps): JSX.Element {
-  const {tab, selectedModuleName, modulesTab} = useParams();
+  const {tab, modulesTab} = useParams();
   const navigate = useNavigate();
   let effectiveTab: TabValue;
-  if (selectedModuleName) {
+  if (modulesTab) {
     effectiveTab = "modules" as TabValue;
   } else if (tab !== undefined) {
     effectiveTab = tab as TabValue;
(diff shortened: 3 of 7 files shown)


src/ExplorerRoutes.tsx
```diff
@@ -31,11 +31,7 @@ export default function ExplorerRoutes() {
         </Route>
         <Route path="/account">
           <Route
-            path=":address/modules/:selectedModuleName/:modulesTab"
-            element={<AccountPage />}
-          />
-          <Route
-            path=":address/modules/:selectedModuleName"
+            path=":address/modules/:modulesTab/:selectedModuleName"
```
jillxuu, line 34: makes sense to me, tho it could be a breaking change if other people used old urls to do something. but i think effect should be minimal.

jillxuu (changes requested):
could you check in the codebase and make sure you update the url for all `/modules` related url redirection? there are couple of them that need to get updated, otherwise, experience's broken.

i.e. go to the txn page, randomly click on a function name, right now its 404.

jillxuu (commented):
code lgtm, one thing i'd like to call out is that i was testing with view function interaction, but txn failed with 400.

0xbe1:
> code lgtm, one thing i'd like to call out is that i was testing with view function interaction, but txn failed with 400.

@jillxuu  can you show me how to reproduce the 400? also are you using devnet?

jillxuu:
Its on devnet and i was using the netlify built urls. Sent from my iPhoneOn Apr 24, 2023, at 10:36 PM, 0xbe1 ***@***.***> wrote:﻿

code lgtm, one thing i'd like to call out is that i was testing with view function interaction, but txn failed with 400.

@jillxuu  can you show me how to reproduce the 400? also are you using devnet?

geekflyer (changes requested):
Let's' rename these module tab URLs to `code|run-entry|run-view` to align more with the terminology in CLI. Calling this `modules/read+write` is kinda confusing. See https://aptos-org.slack.com/archives/C040LV2NWQ4/p1680757174899329?thread_ts=1680740887.034249&cid=C040LV2NWQ4

Also can we add url navigation support? It's kinda annoying that currently if I'm let's say looking at the staking module I have to scroll all the way down whenever I switch between code|run-entry|run-view  to get to the staking module and I can't share a link to a specific function.

Lastly this just some thought - no action required - but from a general UX perspective I think separating run-entry and run-view into separate tabs adds clunkiness. They should be ideally in the same tab (maybe with some optional filtering if necessary).

0xbe1:
@geekflyer thx for the feedback.

> Let's' rename these module tab URLs to code|run-entry|run-view to align more with the terminology in CLI.

in cli we do `aptos move view` vs `aptos move run`. so if we want to align then we probably want to do code|run|view instead? btw not able to access the slack link since i am not in the slack group

> Also can we add url navigation support? It's kinda annoying that currently if I'm let's say looking at the staking module I have to scroll all the way down whenever I switch between code|run-entry|run-view to get to the staking module and I can't share a link to a specific function.

plan to support this in separate pr. step by step.

> separating run-entry and run-view into separate tabs is unnecessary

yep i think we can decide later. currently the code is in one place, so very easy to merge them into one

geekflyer:
> @geekflyer thx for the feedback.
> 
> > Let's' rename these module tab URLs to code|run-entry|run-view to align more with the terminology in CLI.
> 
> in cli we do `aptos move view` vs `aptos move run`. so if we want to align then we probably want to do code|run|view instead? btw not able to access the slack link since i am not in the slack group
> 
> > Also can we add url navigation support? It's kinda annoying that currently if I'm let's say looking at the staking module I have to scroll all the way down whenever I switch between code|run-entry|run-view to get to the staking module and I can't share a link to a specific function.
> 
> plan to support this in separate pr. step by step.
> 
> > separating run-entry and run-view into separate tabs is unnecessary
> 
> yep i think we can decide later. currently the code is in one place, so very easy to merge them into one

On 1.: Ok yeah code|run|view works. I would also suggest to rename the UI labels accordingly.

On 2.: Ok, but please make it a fast follow up. Sometimes it's hard to add these things after the fact then the code is structured in a certain way.

0xbe1:
@geekflyer fixed (1), ok to unblock?

jillxuu (approved):
comments in the scope of this PR are addressed. stamp to unblock.

Following AI is to add url navigation support so that move across tabs will persist the function/module

geekflyer (approved):

PR #486, opened 2023-04-24
upgrade wallet adapter and remove wallet connection drop workaround

geekflyer:
This upgrades wallet adapter to the latest version which includes (https://github.com/aptos-labs/aptos-wallet-adapter/pull/129) which fixes the wallet auto reconnect. Removing the now obsolete window.onload workaround so that the page loads+renders at normal speed again.
Tested locally the wallet connnection flow + refresh a few times and it works as expected.

Also making some minor edits to the tsconfig.json (in particular removing isolatedModules: true) which makes VS Code properly suggest auto imports of other modules which didn't work previously.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
   "dependencies": {
     "@apollo/client": "^3.7.9",
     "@aptos-labs/aptos-names-connector": "^0.0.2",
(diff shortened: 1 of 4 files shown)


jillxuu (approved):
awesome thx for the fix! GP-221

PR #487, opened 2023-04-25
fix data race when fetching delegation pool info and upgrade react query

geekflyer:
This PR fixes a data race in the code that fetches delegation pool info.
Previously on some page refreshes it was sometimes showing incorrect / stale data for delegation pools in the delegation pool table (such as showing random delegation pools as active, even though they're not).
This - I think - should fix that.

The main changes are:
1. update react-query (this is kinda unrelated to the bug, just wanted to upgrade to latest version so that the docs apply etc.)
2. refactor the useGetDelegationNodeInfo hook. The main problem was that this hook was performing a bunch of async side effects, such as fetching data inside useMeme. Because these async requests don't get cancelled when useMemo re-executes, a delayed fetch response was able to set the validator status of a Table Row to an incorrect status. In general performing any side effect (such as setState), but especially an async one (such as fetching data) inside useMemo is a bad idea.
See also the [react docs on useMemo](https://legacy.reactjs.org/docs/hooks-reference.html#:~:text=Remember%20that%20the%20function%20passed%20to%20useMemo%20runs%20during%20rendering.%20Don%E2%80%99t%20do%20anything%20there%20that%20you%20wouldn%E2%80%99t%20normally%20do%20while%20rendering.%20For%20example%2C%20side%20effects%20belong%20in%20useEffect%2C%20not%20useMemo.)

Fyi there are a few 2 or 3 more places that perform a setState inside useMemo in the codebase, which is a side-effect.


jillxuu (approved):
lgtm! thx for the fix. could you also update this hook to use `useQuery` as well in the same pr? https://github.com/aptos-labs/explorer/blob/f2d605fdbc37630428ffc48de475821ddc6c32b4/src/api/hooks/useGetDelegatorStakeInfo.ts#L6-L21

geekflyer:
> lgtm! thx for the fix. could you also update this hook to use `useQuery` as well in the same pr?
> 
> https://github.com/aptos-labs/explorer/blob/f2d605fdbc37630428ffc48de475821ddc6c32b4/src/api/hooks/useGetDelegatorStakeInfo.ts#L6-L21

oh hmm, this got auto merged already. Can you create another PR to make the change to useQuery? makes sense.

PR #492, opened 2023-04-26
[explorer] fix parameterless view function call

0xbe1:
without default values, submitting the form could potentially submit `{..., arguments: undefined}` which fails the view function call. after the change, `{..., arguments: []}` won't fail the view function call.

besides, display the error message if view function call fails
image

### DIFF
--- a/src/pages/Account/Tabs/ModulesTab/Contract.tsx
+++ b/src/pages/Account/Tabs/ModulesTab/Contract.tsx
@@ -1,5 +1,5 @@
 import {Types} from "aptos";
-import {ReactNode, useState} from "react";
+import {ReactNode, useEffect, useState} from "react";
 import Error from "../../Error";
 import {useGetAccountModules} from "../../../../api/hooks/useGetAccountModules";
 import EmptyTabContent from "../../../../components/IndividualPageContent/EmptyTabContent";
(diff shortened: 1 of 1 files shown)


0xbe1:
> saw that react console silently swallowed the error and didn't display as the function result. could we fix that?

just to confirm, do you want me to display function call error in the UI as well?

jillxuu:
> just to confirm, do you want me to display function call error in the UI as well? @jillxuu

yes if there are no other error handling mechanisms in place to notify users.

src/pages/Account/Tabs/ModulesTab/Contract.tsx
```diff
@@ -228,15 +228,20 @@ function ReadContractForm({
 }) {
   const [state] = useGlobalState();
   const [result, setResult] = useState<Types.MoveValue[]>([]);
+  const [errMsg, setErrMsg] = useState<string>();
 
   const onSubmit: SubmitHandler<ContractFormType> = async (data) => {
     const viewRequest: Types.ViewRequest = {
-      function: `${module.address}::${module.name}::${fn.name}`,
+      function: `${module.address}::${module.name}::${fn.name}!`,
```
jillxuu, line 235:
```suggestion
      function: `${module.address}::${module.name}::${fn.name}`,
```

src/pages/Account/Tabs/ModulesTab/Contract.tsx
```diff
@@ -228,15 +228,20 @@ function ReadContractForm({
...
       arguments: data.args,
     };
-    const result = await view(viewRequest, state.network_value);
-    setResult(result);
+    try {
+      const result = await view(viewRequest, state.network_value);
+      setResult(result);
+    } catch (e: any) {
+      setErrMsg(e.message ?? String(e));
+    }
```
jillxuu, line 244: this error msg persist even if i switch function names

0xbe1: @jillxuu can you help me understand why? after switching the function name, i thought the ReadContractForm will re-render since props.fn change, therefore reset all internal states. however it doesn't

jillxuu: it seems like the entire component's not re-rendering when `fn` changed, not sure exactly why. but a potential fix is to add 
```
 useEffect(() => {
    setErrMsg("");
  }, [fn]);
```
to force re-render when `fn` changes.

jillxuu (approved):

Issue #501, opened 2023-05-02
run and view contract issues


0xbe1:
here's a main thread of pending issues on "run and view contract".

## deep link

in http://localhost:3001/account/0x1/modules/view/acl?network=devnet, clicking `exists_at` from the sidebar should change the URL to http://localhost:3001/account/0x1/modules/view/exists_at?network=devnet, etc.

## mobile view

as in https://www.figma.com/file/HoHbGZxCJYfyR3A1PRjvWG/Contract-features---View-code%2C-Read%2C-Write?node-id=0-1&t=5bTHPMQLDhXbQIiu-0

## more styling

make sure implementation matches figma design https://www.figma.com/file/HoHbGZxCJYfyR3A1PRjvWG/Contract-features---View-code%2C-Read%2C-Write?node-id=0-1&t=5bTHPMQLDhXbQIiu-0

PR #502, opened 2023-05-05
Router for Run and View contract

SamuelQZQ:
#501 

Before this change, only the "code" tab on the modules page align the URL with the tab state. 
For example: `http://localhost:3000/account/0x1/modules/code/acl?network=devnet&feature=dev`

After this change, the "run" and "view" tabs will also align the URL with the tab state. But for the "run" and "view" tabs, the sidebar item is a function in a module, instead of the module itself.
So the link would be: `http://localhost:3000/account/0x1/modules/view/${moduleName}/${functionName}?network=devnet&feature=dev`

And when switch between tabs, we will keep the URL state. For example:
-  If the user is looking at the `balance` function in the `coin` module, which is `http://localhost:3000/account/0x1/modules/view/coin/balance?network=devnet&feature=dev`.
-  Then the user switches to the `code` tab. The code of the `coin` module will directly be opened.
- Then the user switches back to the `view` tab. He can still see the `balance` function in the `coin` module.

### DIFF
--- a/src/ExplorerRoutes.tsx
+++ b/src/ExplorerRoutes.tsx
@@ -34,6 +34,10 @@ export default function ExplorerRoutes() {
             path=":address/modules/:modulesTab/:selectedModuleName"
             element={<AccountPage />}
           />
+          <Route
+            path=":address/modules/:modulesTab/:selectedModuleName/:selectedFnName"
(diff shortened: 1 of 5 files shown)


0xbe1:
prefer #503

PR #503, opened 2023-05-05
[Code View&Run] Router for Run and View contract

SamuelQZQ:
#501 

Before this change, only the "code" tab on the modules page align the URL with the tab state. 
For example: `http://localhost:3000/account/0x1/modules/code/acl?network=devnet&feature=dev`

After this change, the "run" and "view" tabs will also align the URL with the tab state. But for the "run" and "view" tabs, the sidebar item is a function in a module, instead of the module itself.
So the link would be: `http://localhost:3000/account/0x1/modules/view/${moduleName}/${functionName}?network=devnet&feature=dev`

And when switch between tabs, we will keep the URL state. For example:
-  If the user is looking at the `balance` function in the `coin` module, which is `http://localhost:3000/account/0x1/modules/view/coin/balance?network=devnet&feature=dev`.
-  Then the user switches to the `code` tab. The code of the `coin` module will directly be opened.

### DIFF
--- a/src/pages/Account/Tabs/ModulesTab/Contract.tsx
+++ b/src/pages/Account/Tabs/ModulesTab/Contract.tsx
@@ -3,6 +3,7 @@ import {ReactNode, useEffect, useState} from "react";
 import Error from "../../Error";
 import {useGetAccountModules} from "../../../../api/hooks/useGetAccountModules";
 import EmptyTabContent from "../../../../components/IndividualPageContent/EmptyTabContent";
+import SidebarItem from "../../Components/SidebarItem";
 import {WalletConnector} from "@aptos-labs/wallet-adapter-mui-design";
 import {useWallet} from "@aptos-labs/wallet-adapter-react";
 import {
@@ -21,10 +22,12 @@ import {
 } from "@mui/material";
 import React from "react";
 import {useForm, SubmitHandler, Controller} from "react-hook-form";
+import {useParams} from "react-router-dom";
 import useSubmitTransaction from "../../../../api/hooks/useSubmitTransaction";
 import {useGlobalState} from "../../../../global-config/GlobalConfig";
 import {view} from "../../../../api";
 import {grey} from "../../../../themes/colors/aptosColorPalette";
+import { useNavigate } from "../../../../routing";
 
 type ContractFormType = {
   typeArgs: string[];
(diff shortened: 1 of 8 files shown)


src/pages/Account/Tabs/ModulesTab/Contract.tsx
```diff
@@ -31,13 +33,13 @@ interface ContractSidebarProps {
...
-  handleClick: (moduleName: string, fnName: string) => void;
+  getLinkToFn(moduleName: string, fnName: string): string;
 }
 
 function Contract({address, isRead}: {address: string; isRead: boolean}) {
   const {data, isLoading, error} = useGetAccountModules(address);
-  const [selectedModuleName, setSelectedModuleName] = useState<string>();
-  const [selectedFnName, setSelectedFnName] = useState<string>();
+  const selectedModuleName = useParams().selectedModuleName ?? "";
+  const selectedFnName = useParams().selectedFnName ?? "";
```
0xbe1, line 42: const { selectedModuleName, selectedFnName } = useParams()

0xbe1 (approved):
lgtm

jillxuu (changes requested):
the url param's not persisted 

https://user-images.githubusercontent.com/121921928/236883820-92268526-61ac-48b8-ba2e-9f6c5cd3bc46.mov

SamuelQZQ:
> the url param's not persisted
> 
>  Screen.Recording.2023-05-08.at.9.53.56.AM.mov

Hi @jillxuu, sorry for the confusion. That's because in a following commit, I removed the `fn` part of the URL. The behavior changed, but I forgot to update the PR description.

I removed the `fn` because the `fn` from the `View` tab is not helpful for the `Code` and `Run` tabs. Currently, we only persist the "module" part in the URL, because that's shared between tabs. It's a trade-off here. I will update the PR description.

jillxuu:
> > the url param's not persisted
> > Screen.Recording.2023-05-08.at.9.53.56.AM.mov
> 
> Hi @jillxuu, sorry for the confusion. That's because in a following commit, I removed the `fn` part of the URL. The behavior changed, but I forgot to update the PR description.
> 
> I removed the `fn` because the `fn` from the `View` tab is not helpful for the `Code` and `Run` tabs. Currently, we only persist the "module" part in the URL, because that's shared between tabs. It's a trade-off here. Which one do you think is better? To keep the `fn` part or remove it?
> 
> I will update the PR description to align with the current behavior.

i believe we should at least check whether the current function exists after switching the tab, if so, persist the function so that experience's consistent, if not, still persist the function in the url so that when tab's switching back, function in url will be reflected if we don't select a new function. 

if after tab switching we select a different function, that new function will then be persisted in the uri instead.

SamuelQZQ:
My last commit adds the functionality to persist the function name in the URL, so that switching between tabs won't lose the state.

And I shared the functionality to jump to the line of the function for the Transaction page and Modules page. To make it consistent, I change the "entry_function" URL query params to the same pathname as the Modules page does.

I noticed a bug about the get function line number Regex, which failed to get the line number of the function below because of the line breaker between type params:
```
 public entry fun two_step_route<
      X, Y, Z, E1, E2,
  >(
      sender: &signer,
      first_dex_type: u8,
      first_pool_type: u64,
      first_is_x_to_y: bool, // first trade uses normal order
      second_dex_type: u8,
      second_pool_type: u64,
      second_is_x_to_y: bool, // second trade uses normal order
      x_in: u64,
      z_min_out: u64,
  ) acquires EventStore, CoinStore, TortugaSigner { }
```

I proposed a fix, please help check.

You can use this link to test:
https://deploy-preview-503--aptos-explorer.netlify.app/account/0xbeaa9e5ef5bee0781476a4adf293aae7dc3a28e9bd79fda89fca7211fb94c80/modules/code/aggregator/two_step_route?network=mainnet&feature=dev

0xbe1 (approved):
lgtm

jillxuu (approved):
thx for addressing these feedbacks!

jillxuu (approved):
thx for addressing these feedbacks!

PR #504, opened 2023-05-06
[Code View&Run] Clear state when function changed

SamuelQZQ:
Clear all of the states in the Form component when `fn` changes. So that we can avoid bugs like:
- Query result isn't cleaned after changing the selected `fn`
- Args are not cleaned after changing the selected `fn`


0xbe1 (approved):
lgtm

PR #512, opened 2023-05-12
[Contract] Add logging

SamuelQZQ:


SamuelQZQ:
Please look #515 instead.

PR #514, opened 2023-05-12
[Contract] hide Copy Code and Expand button when code not available

SamuelQZQ:
image


SamuelQZQ:
Please look #516 instead.

PR #518, opened 2023-05-16
[Contract] Validate the args, make sure it's not empty

SamuelQZQ:
Validate that the args are not empty before submission.

Screenshot:
image

I changed the `theme` because the disabled button text color is not readable
image

### DIFF
--- a/src/pages/Account/Tabs/ModulesTab/Contract.tsx
+++ b/src/pages/Account/Tabs/ModulesTab/Contract.tsx
@@ -319,6 +319,7 @@ function RunContractForm({
           <Box>
             <StyledTooltip
               title="Input arguments cannot be empty."
+              placement="right"
               disableHoverListener={formValid}
             >
               <span>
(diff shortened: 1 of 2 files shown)


0xbe1:
Can we change it so that if args empty the button is disabled

src/pages/Account/Tabs/ModulesTab/Contract.tsx
```diff
@@ -313,21 +313,31 @@ function RunContractForm({
...
+      setFormValid={setFormValid}
       result={
         connected ? (
           <Box>
-            <Button
-              type="submit"
-              disabled={transactionInProcess}
-              variant="contained"
-              sx={{width: "8rem", height: "3rem"}}
+            <StyledTooltip
```
jillxuu, line 320: can we have the tooltip display on the right hand side?

jillxuu (approved):

PR #521, opened 2023-05-17
fix ans name query race condition

jillxuu:
fix ans name query race condition by using `useQuery` instead of `useEffect`
https://aptos-org.slack.com/archives/C040LV2NWQ4/p1683841215530359

### DIFF
--- a/src/api/hooks/useGetANS.ts
+++ b/src/api/hooks/useGetANS.ts
@@ -1,4 +1,4 @@
-import {useEffect, useState} from "react";
+import {useQuery} from "@tanstack/react-query";
 import {NetworkName} from "../../constants";
 import {useGlobalState} from "../../global-config/GlobalConfig";
 import {
@@ -7,6 +7,7 @@ import {
   setLocalStorageWithExpiry,
   truncateAptSuffix,
 } from "../../utils";
+import {ResponseError} from "../client";
 
 const TTL = 60000; // 1 minute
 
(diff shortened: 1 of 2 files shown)


src/api/hooks/useGetANS.ts
```diff
@@ -26,43 +27,53 @@ function getFetchNameUrl(
...
-          }
-          setName(name);
-        }
-      };
-
-      fetchData().catch((error) => {
-        console.error("ERROR!", error, typeof error);
-      });
+  // return empty string if ans name not found to prevent useQuery complaining about undefined return
+  return queryResult.data === "" ? undefined : queryResult.data;
```
geekflyer, line 42: I don't quite get this. Why are we doing this? This also doesn't seem to do what the comment says, aka returning an empty string?

src/api/hooks/useGetANS.ts
```diff
@@ -26,43 +27,53 @@ function getFetchNameUrl(
...
+
+async function genANSName(
+  address: string,
+  shouldCache: boolean,
+  networkName: NetworkName,
+): Promise<string> {
+  const primaryNameUrl = getFetchNameUrl(networkName, address, true);
+
+  if (!primaryNameUrl) {
+    return "";
```
geekflyer, line 53: why not return undefined instead?

src/api/hooks/useGetANS.ts
```diff
@@ -26,43 +27,53 @@ function getFetchNameUrl(
...
+  try {
+    const {name: primaryName} = await fetchJsonResponse(primaryNameUrl);
+
+    if (primaryName) {
+      if (shouldCache) {
+        setLocalStorageWithExpiry(address, primaryName, TTL);
+      }
+      return primaryName;
+    } else {
+      const nameUrl = getFetchNameUrl(networkName, address, false) ?? "";
```
geekflyer, line 65: this seems kinda bad for error noise. We're defaulting to an empty string and then make an http request to empty string which will certainly lead to an error. Let's just return undefined and skip the request part as a result.

src/api/hooks/useGetANS.ts
```diff
@@ -26,43 +27,53 @@ function getFetchNameUrl(
...
+      const {name} = await fetchJsonResponse(nameUrl);
+      if (shouldCache) {
+        setLocalStorageWithExpiry(address, name, TTL);
+      }
+      return name || "";
     }
-  }, [address, state]);
+  } catch (error) {
+    console.error("ERROR!", error, typeof error);
```
geekflyer, line 73: let's make the error more descriptive - aka saying something akin to could not find or fetch ANS name.

geekflyer (changes requested):

src/api/hooks/useGetANS.ts
jillxuu: undefined's not an accepted value to be returned from the query function. pls checkout this link below:

https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#undefined-is-an-illegal-cache-value-for-successful-queries

so instead i have to make it return empty strings if ans name not found.

src/api/hooks/useGetANS.ts
jillxuu: ditto

src/api/hooks/useGetANS.ts
geekflyer: how about `null` then?

geekflyer: regardless the comment above is confusing above the return statement. If you put this comment above the `genANSName` it makes more sense. Also would be good to add the link to the tanstack query docs in the code for full context.
Regardless I think it's better to return null instead of empty string.

geekflyer (changes requested):
see comment about null

src/api/hooks/useGetANS.ts
```diff
@@ -26,43 +27,63 @@ function getFetchNameUrl(
...
+
+      if (!nameUrl) {
+        return null;
+      }
+
+      const {name} = await fetchJsonResponse(nameUrl);
+      if (shouldCache) {
+        setLocalStorageWithExpiry(address, name, TTL);
+      }
+      return name || null;
```
geekflyer, line 76:
```suggestion
      return name ?? null;
```

geekflyer (approved):

PR #540, opened 2023-05-19
migrate from abandoned create-react-app to vite

geekflyer:
This migrates from [create-react-app](https://github.com/facebook/create-react-app) which is kinda abandoned to using https://vitejs.dev/ . For further motivation see https://dev.to/ag2byte/create-react-app-is-officially-dead-h7o .
The main motivation I'm making this change is that the stacktraces in production are pretty unreadable since create-react-app transpiles async functions to older JS syntax even though that's unnecessary with our target browsers and create-react-app is not configurable for this.
But also create-react-app seems abandoned anyways.

I followed mostly this guide for the migration https://www.robinwieruch.de/vite-create-react-app/ .

Bonus: I added some very basic eslint setup - currently only enforcing rules of hooks. Turns out we have a quite number of violations of them in the current codebase.

### Test Plan
Manual testing and ran all pnpm commands.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.1",
   "private": true,
   "packageManager": "pnpm@7.32.5",
+  "type": "module",
   "dependencies": {
     "@apollo/client": "^3.7.9",
     "@aptos-labs/aptos-names-connector": "^0.0.2",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -66,16 +68,18 @@
     "typescript": "^5.0.2"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
+    "start": "vite --port 3000",
```
jillxuu, line 71: hardcoding port number not necessary? when running couple project locally, 3000 will be in conflict anyways

jillxuu (approved):

PR #547, opened 2023-05-26
fix lint error

jillxuu:
TODO: future pr will fix lint warnings

this pr fixes most lint errors by making sure hook order's consistent, way of doing this is to abstract out the conditional check logic: https://github.com/aptos-labs/explorer/actions/runs/5094789130/jobs/9159010587?pr=547


src/pages/DelegatoryValidator/StakeOperationDialog.tsx
```diff
@@ -95,6 +123,28 @@ export default function StakeOperationDialog({
...
+
+  useEffect(() => {
+    async function fetchData() {
+      if (stakeOperation === StakeOperation.STAKE) {
+        const fee = await getAddStakeFee(
+          client,
+          validator!.owner_address,
+          Number(amount).toFixed(8),
+        );
+        setAddStakeFee(fee[0]);
```
geekflyer, line 143: don't have to address it necessarily in this PR but btw this data fetching logic is subject to race conditions and may display stale data.

geekflyer (approved):

src/pages/DelegatoryValidator/StakeOperationDialog.tsx
jillxuu: will address all the useQuery & race conditions in future PRs.

PR #550, opened 2023-05-31
fix lint warning

jillxuu:
```
jillxu@Jills-MacBook-Pro explorer % pnpm lint

> aptos-explorer@0.0.1 lint /Users/jillxu/Documents/GitHub/explorer
> tsc --noEmit && eslint src --ext ts,tsx --report-unused-disable-directives --max-warnings 0
```

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,247 +1,173 @@
-lockfileVersion: '6.0'
+lockfileVersion: 5.4
+
+specifiers:
+  '@apollo/client': ^3.7.9
+  '@aptos-labs/aptos-names-connector': ^0.0.2
+  '@aptos-labs/wallet-adapter-mui-design': ^0.3.1
+  '@aptos-labs/wallet-adapter-react': ^1.0.4
+  '@blocto/aptos-wallet-adapter-plugin': ^0.1.4
+  '@download/blockies': ^1.0.3
+  '@emotion/react': latest
+  '@emotion/styled': latest
+  '@jest/globals': ^29.5.0
+  '@martianwallet/aptos-wallet-adapter': ^0.0.4
+  '@mui/icons-material': ^5.11.11
+  '@mui/material': ^5.11.13
+  '@mui/system': ^5.11.13
+  '@mui/x-date-pickers': ^5.0.20
+  '@nightlylabs/aptos-wallet-adapter-plugin': ^0.2.12
+  '@openblockhq/aptos-wallet-adapter': ^0.1.5
+  '@pontem/wallet-adapter-plugin': ^0.1.4
+  '@rise-wallet/wallet-adapter': ^0.1.2
+  '@sentry/react': ^7.0.0
+  '@sentry/tracing': ^7.0.0
+  '@solana/web3.js': ^1.75.0
+  '@tanstack/react-query': ^4.29.5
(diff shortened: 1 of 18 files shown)


pnpm-lock.yaml
```diff
@@ -1,247 +1,173 @@
-lockfileVersion: '6.0'
+lockfileVersion: 5.4
```
geekflyer, line 1: please update your pnpm version to 8.x and revert this lockfile change

src/api/hooks/useGetSearchResults.ts
```diff
@@ -143,15 +140,9 @@ export default function useGetSearchResults(input: string) {
       }
 
       const resultsList = await Promise.all(promises);
-      const results = resultsList
-        .filter((result): result is SearchResult => !!result)
-        .map((result) => ({
-          ...result,
-          to:
-            result.to !== null
-              ? augmentToWithGlobalSearchParams(result.to)
```
geekflyer, line 152: you sure we can remove this?

src/global-config/network-selection.ts
```diff
@@ -42,35 +42,31 @@ export function useNetworkSelector() {
...
+        (prev) => {
+          const newParams = new URLSearchParams(prev);
+          newParams.set("network", network);
+          return newParams;
+        },
+        {replace},
+      );
+      writeSelectedNetworkToLocalStorage(network);
     },
-    [], // empty [] makes this effect only run once (on mount)
```
geekflyer, line 71: I think this is a case where we most likely want the previous behaviour and only run this effect on initial mount. Perhaps add some `// eslint-ignore <rule-name>` here to stop the linter from complaining about it.

Let's just undo the code change in this file and add this eslint ignore comment

src/pages/Account/Tabs/ModulesTab/Tabs.tsx
```diff
@@ -93,7 +93,7 @@ function ModulesTabs({address}: {address: string}): JSX.Element {
         break;
     }
     eventName && logEvent(eventName);
-  }, [value]);
+  }, [value, logEvent]);
```
geekflyer, line 96: looking at the implementation of useLogEvent it looks like it returns a new function on every rerender and hence would retrigger this effect every time, which is probably not what we want.

src/pages/Validators/Index.tsx
```diff
@@ -54,7 +53,15 @@ export default function ValidatorsPage() {
         },
       );
     }
-  }, []);
```
geekflyer, line 57: might be another case where we want to trigger the event only once / on initial mount? not sure though without digging deeper into the logic

geekflyer (changes requested):

pnpm-lock.yaml
jillxuu: pnpm version is alraedy 8.6.0 but if i run `pnpm i` this file will be generated again with version 5.4

geekflyer: are you sure? that sounds very odd

jillxuu: https://github.com/aptos-labs/explorer/assets/121921928/175cd260-9a72-4402-a7e4-adddffa5ab8e

geekflyer: you forgot the most important part in the video - `pnpm --version` :-P

0xZihan (approved):

PR #554, opened 2023-06-02
[Contract] Fix bool & vector for contract execute

SamuelQZQ:
We can test on this account in testnet: 0x3d097bb505c9e5d8a96e367f371168240025877f6be8d4a88eacaafb709fe5c9.

After this change, we support user input hex or array for `vector<u8>`.

Meanwhile, I figured out that the BCS serialization for `view` function is not necessary, so I removed that code.

`view` function can accept `vector<u8>` as hex. `submit` function can accept `vector<u8>` as an array of strings.

I haven't tested vector type other than `vector<u8>`. I will deploy some contracts to test to see if it's the same. When I figured it out, I will make a new PR to fix other types.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,6 +1,5 @@
 import {BCS, HexString, TxnBuilderTypes, Types} from "aptos";
 import pako from "pako";
-
 /**
  * Helper function for exhaustiveness checks.
  *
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -161,6 +161,20 @@ export function getPublicFunctionLineNumber(
   return 0;
 }
 
+export function encodeInputArgsForViewRequest(type: string, value: string) {
+  if (type.includes("vector")) {
+    // when it's a vector, we support both hex and javascript array format
+    return value.trim().startsWith("0x")
+      ? value.trim()
+      : encodeVectorForViewRequest(type, value);
```
jillxuu, line 169: this might not work well, cuz imagine if i input random string like "aptos" to the view function thats taking vector<u8>, it will end up in `encodeVectorForViewRequest`, which doesn't work in this case. 

Screenshot 2023-06-05 at 3 03 26 PM

jillxuu (changes requested):

src/utils.ts
SamuelQZQ: I pushed a new commit to add some data validaiton.

PR #556, opened 2023-06-05
[sev] only use primary name lookup for validators ans association

jillxuu:
https://aptoslabs.pagerduty.com/incidents/Q3GI53T03EJXT3


angieyth:
accept to unblock. curious why we don't use primary name for everything but only validators?

jillxuu:
> accept to unblock. curious why we don't use primary name for everything but only validators?

cuz other user addresses might still wanna non primary names and we don't wanna disable that functionality for all. just for validators, we wanna be extra cautious to avoid malicious intent.

PR #561, opened 2023-06-14
Updates wallet-adapter and react-number-format

kaw2k:
# Description

The wallet adapter [updated to include deep linking support](https://github.com/aptos-labs/aptos-wallet-adapter/pull/149). This PR updates our version to use the new adapter, as well as fixes a lingering import issue with `react-number-format`. [Here is the upgrade guide](https://s-yadav.github.io/react-number-format/docs/migration/) for `react-number-format` for reference. 

# Test Plan
I smoke tested to make sure the tests in the previous PR work when including the new npm modules:

https://github.com/aptos-labs/explorer/assets/694324/c5718364-246c-4e61-89ce-6332b0f523ac

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   "dependencies": {
     "@apollo/client": "^3.7.9",
     "@aptos-labs/aptos-names-connector": "^0.0.2",
-    "@aptos-labs/wallet-adapter-mui-design": "^0.3.1",
-    "@aptos-labs/wallet-adapter-react": "^1.0.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -68,7 +68,7 @@
     "typescript": "^5.0.2"
   },
   "scripts": {
-    "start": "vite --port 3000",
```
jillxuu, line 71: is this intended?

kaw2k: nope! good catch 😄

jillxuu (approved):
awesome thanks for adding the deeplink mobile functionality to explorer!

PR #563, opened 2023-06-28
Support entry function without signer param

SamuelQZQ:
Some entry function does not have the `signer` param. We should also support that.


0xbe1 (approved):
lgtm

jillxuu (approved):

PR #574, opened 2023-07-24
update dependencies

0xmaayan:
Update `aptos`, `wallet-adapter-react` and `wallet-adapter-mui-design` version to be compatible with latest upgrades.


kent-white (approved):
thank you!

gregnazario (approved):

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
gregnazario, line 1: interesting this went down?

Maybe you need to update your local pnpm?

0xmaayan: nice catch! I used a lower version, and needed to update and that probably messed up with the lock file

PR #586, opened 2023-08-15
Token V2 typings and sdk update

kent-white:
http://localhost:3000/token/0xab7e7754e7a7533edcbf698f1cee998296e438f1c84e5db9d4c74dcafe02b36c/1?network=mainnet

### DIFF
--- a/src/api/hooks/useGetAccountTokens.ts
+++ b/src/api/hooks/useGetAccountTokens.ts
@@ -1,3 +1,8 @@
+import {
+  Current_Token_Datas,
+  Current_Token_Datas_V2,
+  Current_Token_Ownerships_V2,
+} from "aptos";
 import {useGlobalState} from "../../global-config/GlobalConfig";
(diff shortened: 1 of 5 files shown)


0xmaayan (approved):

src/api/hooks/useGetAccountTokens.ts
```diff
@@ -59,7 +64,7 @@ export function useGetTokenData(tokenDataId?: string) {
         return undefined;
       }
       const response = await state.indexer_client?.getTokenData(tokenDataId);
-      return response?.current_token_datas_v2 ?? [];
+      return response?.current_token_datas_v2;
```
kaw2k, line 67: Could be tricky changing the return type of a function that isn't explicitly typed. Looks like all usage of this function accounts for falsey though!

kaw2k (approved):

hardsetting (approved):

PR #593, opened 2023-08-24
Move explorer to fungible assets query

kent-white:
Moving the coin balances query to the new fungible assets query, this will include coin v1 and fungible assets as seen in the account below.

http://localhost:3000/account/0x40b880d668a9d68d7d521bec03b92857363339053c28459fdfd8d3d905ff0379/coins?network=testnet

vs 

the live version missing the FA coin
https://explorer.aptoslabs.com/account/0x40b880d668a9d68d7d521bec03b92857363339053c28459fdfd8d3d905ff0379/coins?network=testnet

### DIFF
--- a/src/pages/Account/Components/CoinsTable.tsx
+++ b/src/pages/Account/Components/CoinsTable.tsx
@@ -22,14 +22,14 @@ function CoinNameCell({coin}: CoinCellProps) {
         textOverflow: "ellipsis",
       }}
     >
-      {coin?.coin_info?.name}
+      {coin?.metadata?.name}
     </GeneralTableCell>
   );
 }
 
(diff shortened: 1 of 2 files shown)


kent-white:
@0xmaayan I believe the sdk doesn't support the fungible assets queries? But when they do, let me know because I want to support types here :)

0xmaayan:
> @0xmaayan I believe the sdk doesn't support the fungible assets queries? But when they do, let me know because I want to support types here :)

Work, work, work, work, work, work
https://github.com/aptos-labs/aptos-core/pull/9747

kent-white:
Will followup this PR with a typed query from the sdk

kaw2k (approved):

PR #596, opened 2023-08-24
Support Option type in Contract Tab

BriungRi:


kaw2k (approved):
Love it, just tested with `run` and `view`.

src/pages/Account/Tabs/ModulesTab/Contract.tsx
```diff
@@ -683,16 +688,18 @@ function ContractForm({
                 <TextField label="signer" disabled fullWidth />
               ))}
             {fnParams.map((param, i) => {
+              // TODO: Need a nice way to differentiate between option and empty string
```
kaw2k, line 691: Talking with @BriungRi , we would probably want a dropdown between some and none

kaw2k:
@kent-white thoughts?

kent-white (approved):
looks good to me.

PR #597, opened 2023-08-26
Normalize address's for transactions

kent-white:
See the amounts shown for this bad address

http://localhost:3000/account/0x0628e72def46bc247908f0b06367bfbf744a50328e429a97632941035ef50a63/transactions?network=mainnet

Future Work:
We can do this better to normalize coming from the node, but I didn't want to try and figure that out today without more testing, so this is a bandaid solution with a little bit of cleanup

### DIFF
--- a/src/pages/Transaction/utils.tsx
+++ b/src/pages/Transaction/utils.tsx
@@ -1,4 +1,5 @@
 import {Types} from "aptos";
+import {normalizeAddress} from "../../utils";
 
 export type TransactionCounterparty = {
   address: string;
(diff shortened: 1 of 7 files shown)


src/pages/Transaction/utils.tsx
```diff
@@ -105,7 +106,7 @@ function getBalanceMap(transaction: Types.Transaction) {
       },
       event: Types.Event,
     ) => {
-      const addr = event.guid.account_address;
```
bchocho, line 108: So this is the only functional change, right?

kent-white: correct

bchocho (approved):
LGTM

PR #616, opened 2023-10-04
Fix world map missing

kent-white:
## Summary

The github json we were using got killed. I found the raw json, so I'm just going to include it in the code

## Test Plan
We back
Screenshot 2023-10-04 at 4 54 40 PM


sherry-x (approved):
saving the world!!!!

PR #626, opened 2023-10-27
[indexer] add indexer data source for transactions

jillxuu:
https://github.com/aptos-labs/explorer/assets/121921928/95289460-4553-45ed-ba36-a2f9ef37d03c

TODO: 
1. need to get transaction payload "type" into the payload: https://github.com/aptos-labs/aptos-indexer-processors/blob/a9e42dc9d80a69e9d02cd5f19c39e0f3f5cfbeb0/rust/processor/src/models/default_models/transactions.rs#L117-L124
2. add type into the data field in changes: we need to add MoveStructTag address to MoveResource
3. add data to table item: use data from proto


public/world.json
```diff
@@ -1 +1,8080 @@
...
\ No newline at end of file
+{
```
kent-white, line 1: Whats going on here?

src/ExplorerRoutes.tsx
```diff
@@ -29,7 +30,16 @@ export default function ExplorerRoutes() {
           <Route path=":txnHashOrVersion" element={<TransactionPage />} />
           <Route path=":txnHashOrVersion/:tab" element={<TransactionPage />} />
         </Route>
-
+        <Route path="/txn_indexer">
```
kent-white, line 33: why is is a separate path? Shouldn't it be able to handle any txn hash or version?

src/pages/TransactionIndexer/Tabs/BalanceChangeTab.tsx
```diff
@@ -0,0 +1,23 @@
...
+    return <EmptyTabContent />;
+  }
+
+  return (
+    <CoinBalanceChangeTable
+      balanceChanges={balanceChanges}
+      transaction={transaction}
+    />
+  );
+}
```
kent-white, line 23: Are all / most of these files just copy and pasted? This is going to make things hard to maintain. why not just modify the existing routes and components to use indexer?

vite.config.ts.timestamp-1696404500675-cd5d5a355e7fd.mjs
```diff
@@ -0,0 +1,19 @@
+// vite.config.ts
+import { defineConfig } from "file:///Users/jillxu/Documents/GitHub/explorer/node_modules/.pnpm/vite@4.3.9_@types+node@18.15.5/node_modules/vite/dist/node/index.js";
```
kent-white, line 2: Seems like things like this shoudn't be committed?

src/pages/TransactionIndexer/Title.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export default function TransactionTitle({transaction}: TransactionTitleProps) {
+  return (
+    <Stack direction="column" spacing={2} marginX={1}>
+      <Typography variant="h3">Transaction</Typography>
+      <TitleHashButton hash={transaction.hash} type={HashType.TRANSACTION} />
+      <TransactionType type={transaction.transaction_type} />
+    </Stack>
+  );
+}
```
kent-white, line 18: For example, this has like no change from the non-indexer version of this component, but we made a whole new file :(

kent-white (changes requested):
back to your queue for some questions on your general approach

src/pages/TransactionIndexer/Tabs/BalanceChangeTab.tsx
jillxuu: to create a fail-safe mechanism, if indexer source goes wrong & goes down, we need to fall back to FN again, so right now to keep things clean and separate, i have a separate folder for the txn from indexer, but, after couple weeks running & testing and things are determined to be good, i'll do the complete migration and remove the old one.

PR #637, opened 2023-12-04
update @welldone-studio/aptos-wallet-adapter package

daoauth:
update @welldone-studio/aptos-wallet-adapter package


gregnazario:
I've updated it

PR #642, opened 2023-12-08
[deps] Update explorer to SDK v2, and update wallet plugins

gregnazario:


src/pages/Account/Tabs/ModulesTab/Contract.tsx
```diff
@@ -299,23 +300,24 @@ function RunContractForm({
...
+            return type === "vector<u8>" && arg.trim().startsWith("0x")
+              ? Array.from(new HexString(arg).toUint8Array()).map((x) =>
+                  x.toString(),
+                )
+              : deserializeVector(arg);
+          } else if (type.startsWith("0x1::option::Option")) {
+            arg ? {vec: [arg]} : undefined;
+          } else return arg;
+        }),
+      },
```
gregnazario, line 320: This unfortunately needs to be reworked, so upgrading to SDK v2 is blocked on this.

SDK V2 in its argument parsing doesn't accept strings for every input, so they'd first have to be parsed into their respective types before passing them on.

PR #643, opened 2023-12-14
fix delegated amount to be the same as voting power

jillxuu:
https://app.incident.io/aptos/incidents/20


ying-w:
confirmed on preview that delegated amount is now same as voting power

ying-w (approved):

PR #647, opened 2024-01-04
[view] Add ledger version to be used as an input for view functions

gregnazario:
I found this useful in debugging a view function for a user.  View functions are allowed to run on any ledger version, so this adds functionality to the explorer to use it.

https://github.com/aptos-labs/explorer/assets/1335270/2cae303a-0019-4919-8daa-acf14228b36c

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -172,9 +172,21 @@ export function getAccountModule(
 export function view(
   request: Types.ViewRequest,
   nodeUrl: string,
+  ledgerVersion?: string,
 ): Promise<Types.MoveValue[]> {
   const client = new AptosClient(nodeUrl);
-  return withResponseError(client.view(request));
+  let parsedVersion = ledgerVersion;
+
+  // Handle non-numbers, to default to the latest ledger version
+  if (typeof ledgerVersion === "string" && isNaN(parseInt(ledgerVersion, 10))) {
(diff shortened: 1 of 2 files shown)


src/api/index.ts
```diff
@@ -172,9 +172,21 @@ export function getAccountModule(
...
+  // Handle non-numbers, to default to the latest ledger version
+  if (typeof ledgerVersion === "string" && isNaN(parseInt(ledgerVersion, 10))) {
+    parsedVersion = undefined;
+  }
+
+  if (ledgerVersion === null || ledgerVersion === "") {
+    return withResponseError(client.view(request, parsedVersion));
+  }
+
+  return withResponseError(client.view(request, ledgerVersion));
```
banool, line 189: What if the ledger version gets beyond the size of number, won't parseInt fail?

gregnazario: The max safe integer for `number` is

9007199254740991

Now, let's say 1M transactions per second

9007199254740991 / 1000000 = 9007199254.740991
9007199254 / 60 seconds / 60 minutes / 24 hours / 365 days = 285.64 years

I think we'll be fine.

kaw2k:
Mind posting either a screenshot or short video demoing the feature?

kaw2k (approved):

PR #648, opened 2024-01-05
Cleanup view/run functions

gregnazario:
- Fixes errors associated with errors that are from invalid ledger versions
- Allows running transactions through mobile in app browsers, even if the experience isn't the greatest
- Stops formatting world.json, so pnpm fmt can be added to CI


kent-white (approved):
good man

PR #654, opened 2024-01-11
[wallets] Statically initialize wallets that don't change for networks

gregnazario:
As mentioned by msafe, reinitializing the wallet adapter, that is expected to have static state can have strange side effects. I've applied this to all wallets integrated with the explorer, that don't take in network specific information.


jillxuu (approved):
lgtm btw wallet adapter ui looks diff from before? Seems like its not using the mui package

gregnazario:
> lgtm btw wallet adapter ui looks diff from before? Seems like its not using the mui package

I don't think I changed that?  will have to check back in on that.

PR #656, opened 2024-01-11
[deps] Upgrade remaining dependencies for Explorer

gregnazario:
This upgrades the remaining dependencies on the explorer (minus moving to SDK v2 in its entirety and tanstack/react-query).

Also, handles `npm audit` concerns with some `pnpm` overrides.


kent-white (approved):
lordy

PR #659, opened 2024-01-12
Set x-indexer-client header for all clients

banool:
Take two.


gregnazario (changes requested):
The preview site is broken due to this change, let's revert the previous one.

gregnazario (approved):

PR #668, opened 2024-01-25
Fix for expiration timestamp formatting issues

kent-white:
Summary
Time stamps are kind of wacky coming back from the node. The request here was to not format expiration_timestamp because its user inputted. The previous logic was interesting and seemingly can come back from the node in a different fashion.

Test plan
http://localhost:3000/txn/369869258?network=mainnet
Screenshot 2024-01-25 at 3 14 52 PM


banool (approved):
Now that's a big date.

PR #670, opened 2024-01-30
add randomnet

zjma:


src/constants.tsx
```diff
@@ -12,6 +12,7 @@ export const networks = {
   local: "http://127.0.0.1:8080/v1",
   previewnet: "https://fullnode-0.previewnet.gcp.aptosdev.com/v1",
   previewnet2: "https://fullnode.previewnet.aptoslabs.com/v1",
+  randomnet: "https://random.aptoslabs.com/v1",
```
sherry-x, line 15: let's use https://fullnode.random.aptoslabs.com/v1 instead, it will be provisioned soon

sherry-x (changes requested):

src/constants.tsx
zjma: fixed

sherry-x (approved):

PR #679, opened 2024-02-26
Use indexer API via API Gateway

banool:
### Description
This is part of the migration to API Gateway. More context here: https://www.notion.so/aptoslabs/Migration-1-Legacy-URL-to-API-Gateway-migration-b007742490684b9bab3debd674504a94.

I found the initial files to change like this:
```
rg -l 'aptoslabs.com'
```

### Test Plan
In the preview you can see that a page that relies on the indexer works, e.g. the tokens page for an account: https://deploy-preview-679--aptos-explorer.netlify.app/account/0x232098630cfad4734812fa37dc18d9b8d59242feabe49259e26318d468a99584/tokens?network=mainnet. It uses the new URL correctly.


gregnazario (approved):
Can't wait until we update it to the new SDK

PR #681, opened 2024-03-01
Support nested vector in explorer

BriungRi:
Use `JSON.parse` to handle parsing `[]` arguments. 

Previously, `[[abc]]` would be parsed as `["[abc]"], but really, it should be `[["abc"]]`.

The noteworthy change here is that quotes must now be used for literals. E.g. `[0xabc]` used to be acceptable but now must be specified as `["0xabc"]`

### DIFF
--- a/src/pages/Account/Tabs/ModulesTab/Contract.tsx
+++ b/src/pages/Account/Tabs/ModulesTab/Contract.tsx
@@ -36,10 +36,7 @@ import {
 import {useLogEventWithBasic} from "../../hooks/useLogEventWithBasic";
 import {ContentCopy} from "@mui/icons-material";
 import StyledTooltip from "../../../../components/StyledTooltip";
-import {
-  deserializeVector,
-  encodeInputArgsForViewRequest,
-} from "../../../../utils";
+import {encodeInputArgsForViewRequest} from "../../../../utils";
 
 type ContractFormType = {
   typeArgs: string[];
@@ -282,7 +279,7 @@ function RunContractForm({
             ? Array.from(new HexString(arg).toUint8Array()).map((x) =>
                 x.toString(),
               )
-            : deserializeVector(arg);
+            : JSON.parse(arg);
         } else if (type.startsWith("0x1::option::Option")) {
           arg ? {vec: [arg]} : undefined;
         } else return arg;
--- a/src/pages/Account/Tabs/ModulesTab/Contract.tsx
+++ b/src/pages/Account/Tabs/ModulesTab/Contract.tsx
@@ -277,6 +277,7 @@ function RunContractForm({
         const type = fnParams[i];
         if (type.includes("vector")) {
           // when it's a vector<u8>, we support both hex and javascript array format
+          console.log(type, arg);
           return type === "vector<u8>" && arg.trim().startsWith("0x")
             ? Array.from(new HexString(arg).toUint8Array()).map((x) =>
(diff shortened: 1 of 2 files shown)


src/pages/Account/Tabs/ModulesTab/Contract.tsx
```diff
@@ -277,6 +277,7 @@ function RunContractForm({
         const type = fnParams[i];
         if (type.includes("vector")) {
           // when it's a vector<u8>, we support both hex and javascript array format
+          console.log(type, arg);
```
gregnazario, line 280: Remove console.log?

src/utils.ts
```diff
@@ -180,16 +180,36 @@ export function encodeInputArgsForViewRequest(type: string, value: string) {
...
-  if (result[0] === "[" && result[result.length - 1] === "]") {
-    result = result.slice(1, -1);
+// Also works nested [[1, 2], [3, 4]] to [["1", "2"], ["3", "4"]]
+// TODO: Need to type this recursively
+export function deserializeVector(vectorString: string): any {
+  const trimmed = vectorString.trim();
+  let inside = "";
+  if (isArrayStr(trimmed)) {
+    inside = trimmed.slice(1, -1);
   }
```
gregnazario: What's this doing?

Can we put a comment above it?

src/utils.ts
```diff
@@ -180,16 +180,36 @@ export function encodeInputArgsForViewRequest(type: string, value: string) {
 }
 
 // Deserialize "[1,2,3]" or "1,2,3" to ["1", "2", "3"]
-export function deserializeVector(vectorString: string): string[] {
-  let result = vectorString.trim();
-  if (result[0] === "[" && result[result.length - 1] === "]") {
-    result = result.slice(1, -1);
+// Also works nested [[1, 2], [3, 4]] to [["1", "2"], ["3", "4"]]
+// TODO: Need to type this recursively
```
gregnazario: Isn't this recursive now?

src/pages/Account/Tabs/ModulesTab/Contract.tsx
```diff
@@ -287,6 +288,7 @@ function RunContractForm({
         } else return arg;
       }),
     };
+    console.log(payload);
```
kent-white, line 291: logss

kent-white (commented):
logs, otherwise this looks good :)

src/utils.ts
```diff
@@ -180,16 +180,36 @@ export function encodeInputArgsForViewRequest(type: string, value: string) {
...
+    let build = "";
+    for (let i = 0; i < inside.length; i++) {
+      if (inside[i] === "]") {
+        innerArrayStrs.push(build);
+        build = "";
+      } else {
+        build += inside[i];
+      }
+    }
+    return innerArrayStrs.map(deserializeVector);
```
gregnazario, line 205: This code will have the bracket at the beginning?

But, I guess you call it recursively?

I think this misses, can we make a unit test, I suspect that:

```
[[blah,blah],[blah,blah]]
```
Will get to:
```
[blah, blah], [blah, blah]
```
Then 
```
innerArrayStrs.push('[blah, blah')
innerArrayStrs.push(',[blah, blah')
```

Then do that recursively which will return two items

```
'[blah', 'blah'
'','[blah', 'blah'
```

kent-white (approved):

PR #682, opened 2024-03-02
[account] Allow nonexistent account with tokens

gregnazario:
This adds two functionalities:
1. View tokens for an address if it's not an account or object
2. Add links for token and collection objects directly on the token page for tokenv2.

### DIFF
--- a/src/pages/Token/Tabs/OverviewTab.tsx
+++ b/src/pages/Token/Tabs/OverviewTab.tsx
@@ -1,6 +1,6 @@
 import {useParams} from "react-router-dom";
 import {Box, Stack, Typography} from "@mui/material";
-import React, {useState} from "react";
+import React, {Fragment, useState} from "react";
 import HashButton, {HashType} from "../../../components/HashButton";
 import ContentBox from "../../../components/IndividualPageContent/ContentBox";
 import ContentRow from "../../../components/IndividualPageContent/ContentRow";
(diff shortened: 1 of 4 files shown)


jillxuu:
can you share an example address with it?

gregnazario:
> can you share an example address with it?

https://deploy-preview-682--aptos-explorer.netlify.app/account/0x000000000000000000000000000000000000000000000000000000000000beef/tokens?network=testnet

We'll need to do a little more work here where:
1. addresses with tokens are recognized, even if they don't have an account created

Right now, if you have the URL it'll work.

Here's an example of a token and the links to collection and token:
https://deploy-preview-682--aptos-explorer.netlify.app/token/0xe8b4fd35e3219de7a16ab16aba50735c2139769640a3b43645badcd6070e8261/0?network=testnet

src/pages/Token/Tabs/OverviewTab.tsx
```diff
@@ -79,6 +79,31 @@ export default function OverviewTab({data}: OverviewTabProps) {
...
+                <HashButton
+                  hash={data?.token_data_id ?? ""}
+                  type={HashType.OBJECT}
+                />
+              }
+            />
+          </Fragment>
+        ) : (
+          <Fragment></Fragment>
+        )}
```
jillxuu, line 106:
```suggestion
        )}
```

jillxuu (approved):
overall lgtm! this could be incremental efforts, next steps could be: if account not found, but token exist, just get rid of all the other tabs in the account, only show tokens, which should prevent the long loading issue in this case.

PR #685, opened 2024-03-12
fix double negation in the description

igor-aptos:
we cannot verify the code, not that we have no option but to verify it :P


ying-w (approved):
fyi @jillxuu

gregnazario (approved):

PR #691, opened 2024-03-29
[modules] Handle packages without modules

gregnazario:
When there are no modules, it crashes trying to access values that don't exist.  Now, it'll skip it accordingly

Example: https://deploy-preview-691--aptos-explorer.netlify.app/object/0xbd27b6ddb1b58075df6f7199bb68506e0fd20b142171e6b2b2c0ea232773f5aa/modules?network=testnet


0xaptosj:
merged #692 i should have checked explorer channel before trying to fix this issue 🤦

PR #706, opened 2024-04-25
Redirect to object page if it is not an account

BriungRi:
1. Disable retries for loading data. The default is 3 retries, but this results in the page hanging for a long time. Generally, retrying is not necessary
2. If there is object data but no account data, redirect to the object page

### DIFF
--- a/src/pages/Account/Index.tsx
+++ b/src/pages/Account/Index.tsx
@@ -75,8 +75,7 @@ export default function AccountPage({isObject = false}: AccountPageProps) {
     // If we are on the account page, we might be loading an object. This
     // handler will redirect to the object page if no account exists but an
     // object does.
-    if (isObject) return;
-    if (!isLoading) {
+    if (!isObject && !isLoading) {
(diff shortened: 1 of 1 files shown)


src/pages/Account/Index.tsx
```diff
@@ -48,19 +49,40 @@ export function accountPagePath(isObject: boolean) {
...
+
+  const isLoading = objectIsLoading || accountIsLoading;
+  const data = isObject ? objectData : accountData;
+  const error = isObject ? objectError : accountError;
+
+  useEffect(() => {
+    // If we are on the account page, we might be loading an object. This
+    // handler will redirect to the object page if no account exists but an
+    // object does.
+    if (isObject) return;
```
kaw2k, line 78: `return` in `useEffect` is used for cleanup functions. It might work without errors, but not guaranteed in future updates. Better to do something like:
```tsx
if (!isObject && !isLoading) { ... }
```

kaw2k (approved):

PR #709, opened 2024-04-29
[wallets] Add bitget wallet to dropdown

gregnazario:
Last PR missed adding to dropdown

### DIFF
--- a/src/pages/layout/index.tsx
+++ b/src/pages/layout/index.tsx
@@ -14,6 +14,7 @@ import {
   AptosWalletAdapterProvider,
   NetworkName,
 } from "@aptos-labs/wallet-adapter-react";
+import {BitgetWallet} from "@bitget-wallet/aptos-wallet-adapter";
 import {BloctoWallet} from "@blocto/aptos-wallet-adapter-plugin";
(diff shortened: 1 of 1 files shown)


src/pages/layout/index.tsx
```diff
@@ -52,6 +54,7 @@ function walletsForNetwork(network: string) {
     okxWallet,
     martianWallet,
```
davidiw, line 55: probably push martian down due to lack of activity
fewcha is still actively developed so put bitget below?

src/pages/layout/index.tsx
```diff
@@ -64,7 +67,7 @@ function walletsForNetwork(network: string) {
     openBlockWallet,
     tokenPocketWallet,
     trustWallet,
-    welldoneWallet,
+    welldoneWallet
```
davidiw, line 70: I thought it was json style to not have trailing commas?

davidiw (approved):

src/pages/layout/index.tsx
gregnazario: Except that it's Javascript and not JSON.  So you can.

Issue #714, opened 2024-05-04
[Bug] Account > Coins Tab is not working in some cases


teenu4:
`coin.metadata` is null and error breaks the whole page. Example: https://explorer.aptoslabs.com/account/0xe24483420ed1579837f27f3c849a10b240363a73e8f4fc551da7a2d3e56f7b13/coins?network=mainnet

<img width="1013" alt="Screenshot 2024-05-04 at 11 18 32 AM" src="https://github.com/aptos-labs/explorer/assets/16734597/b440f749-136e-4f0c-8619-6ab2e5c7adf4">

PR #715, opened 2024-05-04
Fix null metadata error for Account > Coins Tab

teenu4:
Fixes issue #714 
Screenshot 2024-05-04 at 11 19 12 AM


teenu4:
Looks like the team is already working on this in #713

PR #716, opened 2024-05-06
Fixes account page loading state issue

kaw2k:
The issue was in `useGetAccountResource`. For some reason splatting `options` wasn't working (?????). Tested it a few times, the object was coming through props, but the hook wasn't respecting it. Ended up passing it directly. In the process I upgraded `react-query` from 4 to 5 for better maintenance. 

After / Before Demo
https://github.com/aptos-labs/explorer/assets/694324/9d6586ad-caba-481b-a122-48ba1e09dcc5

### DIFF
--- a/src/api/hooks/useGetAccount.ts
+++ b/src/api/hooks/useGetAccount.ts
@@ -10,11 +10,11 @@ export function useGetAccount(
 ) {
   const [state] = useGlobalState();
 
-  const result = useQuery<Types.AccountData, ResponseError>(
-    ["account", {address}, state.network_value],
-    () => getAccount({address}, state.network_value),
-    options,
-  );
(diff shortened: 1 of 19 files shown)


src/api/hooks/useGetAccount.ts
```diff
@@ -10,11 +10,11 @@ export function useGetAccount(
 ) {
   const [state] = useGlobalState();
 
-  const result = useQuery<Types.AccountData, ResponseError>(
-    ["account", {address}, state.network_value],
-    () => getAccount({address}, state.network_value),
-    options,
-  );
```
igor-p, line 17: I didn't even know this was possible

igor-p (approved):
Looks good!

PR #717, opened 2024-05-07
Upgrade to latest wallet adapter

0xmaayan:
In preparation to Petra new release which introduces Petra to support only the new wallet standard, this PR upgrades Explorer to the latest wallet adapter which supports the new wallet standard (AIP-62).

- I changed the `submitTransaction` function to be compatible with the new adapter and the new ts sdk
- I updated the `Run` and `Stake` functions Explorer supports to be compatible with the new adapter and the new ts sdk
- Removed Nightly wallet plugin as it is a new standard compatible wallet
- Tested with Nightly wallet (which is a new standard compatible wallet) as well with Petra (which is a legacy standard compatible wallet)
- ANSConnector? does Explorer even use it? looks like we only show it if we are on dev mode https://github.com/aptos-labs/explorer/blob/main/src/pages/Account/Components/AptosNamesBanner.tsx#L37. I commented it out for now as the connector package should be updated also to the new sdk and adapter versions

Other than what I tested, anything else we should look at?


kaw2k:
Is this only for signing transactions? If this touches views as well, there are other places needing updates.

0xmaayan:
> Is this only for signing transactions? If this touches views as well, there are other places needing updates.

I think views are using the SDK directly, and that still uses the legacy and old sdk

kaw2k (approved):

PR #724, opened 2024-05-10
Improves search bar

kaw2k:
# I highly recommend you visit the preview deployment and play with the searchbar

This PR tweaks the searchbar:
- Adds loading state that appears as soon as you start typing, goes away after results have loaded
- Only shows options when loading is done, no more "No Results" while a search is active
- Pressing enter auto-selects the first entry, no more need to press down then enter
- Searching for subdomains now works
  - NOTE: There is an issue with fetching expired subdomains with the SDK. When I get to fixing it, this will work without need for an update. 
- Previously there was an "X" on the far right of the search bar when you typed to clear the input. For the life of me I couldn't find out how to have that _and_ a loading indicator. Advice welcome. Or we can just begone with it. 

https://github.com/aptos-labs/explorer/assets/694324/46b4835f-5c36-4b84-b958-002cc9dd0887


package.json
```diff
@@ -73,6 +73,7 @@
   },
   "scripts": {
     "start": "vite --port 3000",
+    "dev": "vite --port 3000",
```
maykonmichel, line 76: Are they supposed to be the same command?

maykonmichel (approved):
SHIP IT!

PR #729, opened 2024-05-18
Code splitting based on routes

rs-8:
Hello!
Introduce code splitting based on routes

Before:
Screenshot 2024-05-18 at 20 22 56
Screenshot 2024-05-18 at 20 21 26

After:
Screenshot 2024-05-18 at 15 30 26
Screenshot 2024-05-18 at 20 24 14

Thanks!

### DIFF
--- a/src/pages/layout/index.tsx
+++ b/src/pages/layout/index.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, {Suspense} from "react";
 import CssBaseline from "@mui/material/CssBaseline";
 import Box from "@mui/material/Box";
 import Container from "@mui/material/Container";
@@ -133,7 +133,7 @@ export default function ExplorerLayout({children}: LayoutProps) {
             >
               <Header />
               <Container maxWidth="xl" sx={{flexGrow: 4, paddingTop: "2rem"}}>
-                {children}
+                <Suspense>{children}</Suspense>
               </Container>
               <Footer />
             </Box>
--- a/src/pages/layout/index.tsx
+++ b/src/pages/layout/index.tsx
@@ -4,6 +4,7 @@ import Box from "@mui/material/Box";
 import Container from "@mui/material/Container";
 import Header from "./Header";
 import Footer from "./Footer";
+import {Fallback} from "./Fallback";
 import {
   GlobalStateProvider,
   useGlobalState,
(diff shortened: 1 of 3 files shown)


src/pages/layout/index.tsx
```diff
@@ -133,7 +133,7 @@ export default function ExplorerLayout({children}: LayoutProps) {
             >
               <Header />
               <Container maxWidth="xl" sx={{flexGrow: 4, paddingTop: "2rem"}}>
-                {children}
+                <Suspense>{children}</Suspense>
```
blakezimmerman, line 136: Can we add a `fallback` here so that something is shown while the page chunk loads?

rs-8: Yes, of course, but I saw that in the case of a fast network we get flickering if we use CircularProgress for all pages, another option is to make a individual skeleton for each page (difficult), so I left it without a fallback


https://github.com/aptos-labs/explorer/assets/26749651/a7a82768-18b2-4e5d-9250-7274db5ce76d

rs-8: I also see that most of the application is loaded gradually without loaders, and in this approach it looks better to me, I could be wrong

blakezimmerman: Yeah, you definitely raise a good point regarding the quick flash of a loading spinner. One approach you could take is to write a fallback component that only shows the spinner after a timeout has elapsed. The following is just an example and you may want to tweak the timeout based on what feels right.

```tsx
function Fallback() {
  const [fallback, setFallback] = useState<ReactNode | null>(null);

  useEffect(() => {
    const timeout = window.setTimeout(() => setFallback(<CircularProgress />), 150);
    return () => {
      window.clearTimeout(timeout);
    };
  }, []);
  
  return fallback;
}
```

rs-8: Added a fallback with a timeout, good solution, thanks for the advice

blakezimmerman (approved):

rs-8:
@blakezimmerman 
Hello!
Can we merge pr?

blakezimmerman:
> @blakezimmerman Hello! Can we merge pr?

Yeah, I think this is ready to go.
@kent-white Should we merge it?

kent-white:
@rs-8 great job!

PR #731, opened 2024-05-21
Update block timestamp

ying-w:
I'm going off the definitions from here: https://aptos.dev/concepts/blockchain/#proposing-the-block

Would it be accurate to say that the timestamp is right before step 8? (Consensus → Execution)


ying-w:
Actually it's step 6. Consensus → Mempool

igor-aptos (approved):

PR #732, opened 2024-05-22
Use staging API Gateway for devnet and testnet

banool:
Internal context: https://www.notion.so/aptoslabs/Route-more-traffic-through-staging-8ac50da7dfcf490fb329b14d9359c6ee.

In short, we want to route more traffic through the staging API Gateway so it is a representative test environment. The explorer is an important but not critical product and a good candidate for this initiative.


kaw2k (approved):
Fine by me, @kent-white not sure if you have any opinions.

PR #766, opened 2024-07-01
Adds fungible assets balance changes

kaw2k:
Went down two separate rabbit holes, took way longer than needed 😓 

Tested with transaction: 0x2ab4b504b1e356ae5539b80f88c38aac0685df5884a7d08497512919d25eac67

image

### DIFF
--- a/src/pages/Transaction/utils.tsx
+++ b/src/pages/Transaction/utils.tsx
@@ -85,7 +85,6 @@ type ChangeData = {
 export type BalanceChange = {
   address: string;
   amount: bigint;
-  amountAfter: string;
   asset: {
     decimals: number;
     symbol: string;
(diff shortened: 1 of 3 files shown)


src/pages/Transaction/utils.tsx
```diff
@@ -178,36 +178,70 @@ function isAptEvent(event: Types.Event, transaction: Types.Transaction) {
...
-      address: key,
-      amount: value.amount,
-      amountAfter: value.amountAfter,
-    });
-  });
+  const balanceChanges: BalanceChange[] =
+    data?.fungible_asset_activities.map((a) => ({
+      address: a.owner_address,
+      amount: BigInt(a.amount),
+      amountAfter: a.amount.toString(),
```
kent-white, line 233: These two are the same?

kent-white (approved):

src/pages/Transaction/utils.tsx
kaw2k: Nope! This is wrong. Thinking back, we decided to remove one column (I think). Was it the after column?

kent-white: yea i think so

PR #780, opened 2024-07-25
fix balance change tab on txn to skip null amount

jillxuu:
example txn to test against: 0x3940661171bdd12eb2eb97579bff5534bacf7f204b6ccf25e87e85c4621d5331


bv-aptos (approved):
LGTM!

PR #784, opened 2024-08-20
Batch delegation nodes API call

BriungRi:
This PR attempts to fix the too many requests issue in the Delegation Nodes tab. Uses an immutable Move helper: https://explorer.aptoslabs.com/account/0xafb5ec0d4cc58dbb1034c2a6d977919f50bc2936dc478274d64c08491c119c4a/modules/view/helpers/pool_address_info?network=testnet

Also fixes the "To many requests" copy

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -304,3 +304,16 @@ export async function getValidatorState(
   };
   return withResponseError(client.view(payload));
 }
+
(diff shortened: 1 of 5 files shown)


gregnazario (approved):
Should definitely cut down on the calls.  We should in the future get this helper probably in the framework

PR #787, opened 2024-09-11
Fixes petra names not showing up in search

kaw2k:
## UPDATED
This PR updates the search logic to always take you to the target address of the name in question, and if that isn't present it tries the owner address of the name. Previously it would take two searches, first to get the name, then the primary name of the owner of the original name. 

Screenshot 2024-10-07 at 11 22 27 AM
Screenshot 2024-10-07 at 11 22 34 AM

## DEPRECATED
Petra names were not showing up in the search bar. This was a combination of issues:
1. An existing issue with the indexer and ans contract
2. We were using an outdated ANS query that was held on to by the TS SDK
3. We had not appended `.apt` to the search text before passing it to the TS SDK.

NOTE: This won't work until `1` gets resolved.

### DIFF
--- a/src/pages/layout/Search/Index.tsx
+++ b/src/pages/layout/Search/Index.tsx
@@ -73,8 +73,8 @@ export default function HeaderSearch() {
 
     const promises = [];
 
-    const isAnsName =
-      searchText.endsWith(".apt") || searchText.endsWith(".petra");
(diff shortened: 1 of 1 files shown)


gregnazario:
@kaw2k can this be merged?

gregnazario (approved):

PR #791, opened 2024-09-25
Handles potentially null states for coin balance changes

kaw2k:
@ying-w found a transaction with both null `owner_address` as well as associated `asset_type`. This PR adds null checking and an empty state for that case. Note the asset type on the second row is gone for the amount. Also, the address on the far left is absent. Previously this would just crash the page. 

image

### DIFF
--- a/src/pages/Transaction/utils.tsx
+++ b/src/pages/Transaction/utils.tsx
@@ -235,8 +235,8 @@ export function useTransactionBalanceChanges(txn_version: string) {
             ? BigInt(-a.amount)
             : BigInt(a.amount),
         asset: {
-          decimals: a.metadata.decimals,
-          symbol: a.metadata.symbol,
+          decimals: a.metadata?.decimals,
+          symbol: a.metadata?.symbol,
         },
       })) ?? [];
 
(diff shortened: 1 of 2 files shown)


src/pages/Transaction/utils.tsx
```diff
@@ -235,8 +235,8 @@ export function useTransactionBalanceChanges(txn_version: string) {
             ? BigInt(-a.amount)
             : BigInt(a.amount),
         asset: {
-          decimals: a.metadata.decimals,
-          symbol: a.metadata.symbol,
+          decimals: a.metadata?.decimals,
```
chiumax, line 238: this causes the page not to crash anymore?

chiumax (approved):

gregnazario (approved):

src/pages/Transaction/utils.tsx
kaw2k: Yes, the `metadata` object was null, and there were no error boundaries, so the entire page crashed.

ying-w:
for context, null states can occur when object is deleted in that transaction

Issue #778, opened 2024-07-24
Gas refund and net gas fee should be shown in overview


ying-w:
Currently, the transaction overview page shows gas units, gas fee, and gas unit price.

However, in situations where multiple resources or table items are deleted, there should be a [large storage refund](https://aptos.dev/en/network/blockchain/gas-txn-fee#calculating-storage-deletion-refund)

For these transactions, the net gas should be negative due to the refund. Please change the UX on explorer so that the refund is shown and also net gas is shown

Example: https://explorer.aptoslabs.com/txn/610075380/userTxnOverview?network=mainnet

- 29 gas units
- 150 octa per gas unit
- 4350 octa of gas used
- 86080 octa storage refund
- -81730 octa net gas

gregnazario:
This looks like it's missing in the FA table handling for storage refunds

PR #809, opened 2024-10-07
[gas] Fix storage fee on main gas page

gregnazario:
Now, gas fee is still the gas charged, but there is a new field that tells the net gas changes.  It additionally makes the balance changes more user friendly.

Before:
https://explorer.aptoslabs.com/txn/610075380/userTxnOverview?network=mainnet
https://explorer.aptoslabs.com/txn/610075380/balanceChange?network=mainnet
https://explorer.aptoslabs.com/txn/1770359587/userTxnOverview?network=mainnet

After:
...

Resolves: https://github.com/aptos-labs/explorer/issues/778


src/pages/Transactions/TransactionsTable.tsx
```diff
@@ -191,6 +191,8 @@ function TransactionAmountGasCell({
               <GasFeeValue
                 gasUsed={transaction.gas_used}
                 gasUnitPrice={transaction.gas_unit_price}
+                transactionData={transaction}
+                netGasCost={true}
```
kaw2k, line 195: nit: you don't need the `={true}`

src/pages/Transaction/utils.tsx
```diff
@@ -227,24 +231,68 @@ export function useTransactionBalanceChanges(txn_version: string) {
...
+        amount: convertAmount(a),
+        type: convertType(a),
         asset: {
           decimals: a.metadata?.decimals,
           symbol: a.metadata?.symbol,
         },
       })) ?? [];
 
+  // Find gas fee and add a storage refund event
+  const gasFeeEvent = data?.fungible_asset_activities.find((a) =>
```
kaw2k, line 281: Is StorageRefund an event we are creating? Do we need to account for it in this check?

src/pages/Transaction/Tabs/UserTransactionOverviewTab.tsx
```diff
@@ -151,10 +151,37 @@ export default function UserTransactionOverviewTab({
...
         />
+        <ContentRow
+          title="Storage Refund:"
+          value={
+            <GasFeeValue
+              gasUsed={transactionData.gas_used}
+              gasUnitPrice={transactionData.gas_unit_price}
+              showGasUsed
+              transactionData={transactionData}
+              storageRefund={true}
```
kaw2k, line 167: nit: drop `={true}`, same on line 183

src/components/IndividualPageContent/ContentValue/GasFeeValue.tsx
```diff
@@ -2,30 +2,90 @@ import React from "react";
...
+            {" ("}
+            <GasValue gas={gasUsed} />
+            {")"}
+          </span>
+        )}
+      </>
+    );
+  }
+
+  return getGasFee();
```
kaw2k, line 90: Standard react style is to have your JSX in your return, preferably not nested in a function invocation. If you need a function invocation, using something like `useMemo` would help. I would recommend something like this:  

```tsx
  const feeStatement = transactionData.events?.find(
      (e) => e.type === "0x1::transaction_fee::FeeStatement",
    );

    if (!feeStatement) {
      return (
        <>
        <APTCurrencyValue
          amount={(BigInt(gasUnitPrice) * BigInt(gasUsed)).toString()}
        />
        {showGasUsed === true && (
          <span style={{color: grey[450]}}>
            {" ("}
            <GasValue gas={gasUsed} />
            {")"}
          </span>
        )}
      </>
      );
    }

    const execution_gas = feeStatement.data.execution_gas_units;
    const io_gas = feeStatement.data.io_gas_units;
    const storage_gas = feeStatement.data.storage_fee_octas;
    const storage_refund = feeStatement.data.storage_fee_refund_octas;

    const gross_gas_units =
      BigInt(execution_gas) + BigInt(io_gas) + BigInt(storage_gas);
    const gross_gas_cost = gross_gas_units * BigInt(gasUnitPrice);
    const net_gas_cost = gross_gas_cost - BigInt(storage_refund);

    if (storageRefund) {
      return (
        <>
          <APTCurrencyValue amount={storage_refund.toString()} />
        </>
      );
    }

    if (netGasCost) {
      return (
        <>
          {net_gas_cost < 0 ? "-" : ""}
          <APTCurrencyValue amount={net_gas_cost.toString()} />
        </>
      );
    }

    return (
      <>
        <APTCurrencyValue amount={gross_gas_cost.toString()} />
        {showGasUsed === true && (
          <span style={{color: grey[450]}}>
            {" ("}
            <GasValue gas={gross_gas_units.toString()} />
            {")"}
          </span>
        )}
      </>
    );
```

src/pages/Transaction/utils.tsx
gregnazario: It's not a real event, I made one up so the UI looks nice.

Gas fee event is also not a real one, but it's created on the indexer

kaw2k (approved):

PR #811, opened 2024-10-07
[search] Add ability to search for addresses that own objects without…

gregnazario:
… account or object status

Example of an address without an account or object

http://localhost:3000/account/0xe7be097a90c18f6bdd53efe0e74bf34393cac2f0ae941523ea196a47b6859edb?network=mainnet

This additionally hides duplicate rows (e.g. object and account).  Account takes precedence, then object, then deleted object, then address.


src/pages/layout/Search/Index.tsx
```diff
@@ -201,15 +217,50 @@ export default function HeaderSearch() {
...
-    const results = resultsList
-      .filter((result) =>
-        wasFound ? !result?.label?.startsWith("Deleted Object") : true,
-      )
+    const isPossiblyAddress = resultsList.find((result) =>
+      result?.label?.startsWith("Address"),
+    );
+
+    let filteredResults: any[] = [];
+    let result: any;
```
kaw2k, line 234: Any way we can remove these `any`s?

src/pages/layout/Search/Index.tsx
```diff
@@ -201,15 +217,50 @@ export default function HeaderSearch() {
...
+      const result = resultsList.find((result) =>
+        result?.label?.startsWith("Address"),
+      );
+      filteredResults = [result];
+    } else {
+      filteredResults = resultsList;
+    }
+    if (filteredResults === undefined || filteredResults === null) {
+      filteredResults = result ? [] : resultsList;
+    }
```
kaw2k, line 260: `Array.find` returns us the value or `null`, so we are doubling our work here. We can simplify it a bit by renaming the predicates to reflect the item that is potentially be found, then test the existence of those items. You don't need the `switch` style check here, I just find it sometimes cleaner than deep `if else` chains as it implies there will ever only be one match. 

```tsx
  const resultsList = await Promise.all(promises);

  const foundAccount = resultsList.find((r) => r?.label?.startsWith('Account'));
  const foundObject = resultsList.find((r) => r?.label?.startsWith('Object'));
  const foundDeletedObject = resultsList.find((r) => r?.label?.startsWith('Deleted Object'));
  const foundPossibleAddress = resultsList.find((r) => r?.label?.startsWith('Address'));

  // Something besides any
  let filteredResults: any[] = [];
  let result: any;
  
  switch (true) { 
    case Boolean(foundAccount): {
      filteredResults = [foundAccount];
      break;
    }
    case Boolean(foundObject): {
      filteredResults = [foundObject];
      break;
    }
    case Boolean(foundDeletedObject): {
      filteredResults = [foundDeletedObject];
      break;
    }
    case Boolean(foundPossibleAddress): {
      filteredResults = [foundPossibleAddress];
      break;
    }
    default: {
      filteredResults = resultsList;
    }
}
```

gregnazario: Okay, this switch is sorcery to me... switch true and then the condition

kaw2k (approved):

PR #820, opened 2024-10-10
display package metadata

0xaptosj:
Closes https://github.com/aptos-labs/explorer/issues/949


gregnazario:
Let's put this data under package, rather than under every module

Screenshot 2024-10-15 at 3 16 16 PM

gregnazario (approved):

PR #821, opened 2024-10-10
Don't crash if txn doesn't have function

CapCap:
Avoid crashing for multisig execution functions
 
image


ying-w:
Similar to how we show SCRIPT as entryfn when a script is submitted, I would have preference for entryfn to say MULTISIG when we have a multisig_payload type with NULL payload 

ex. https://fullnode.mainnet.aptoslabs.com/v1/transactions/by_version/346773297

gregnazario (approved):
Good placeholder.  Let's reevaluate in a follow up issue

PR #824, opened 2024-10-16
[gas] Fix gas used units to use the proper values

gregnazario:


ying-w (approved):
storage is in octa whereas io and execution are in units (that is then multiplied by gas unit price to get octa)

Issue #826, opened 2024-10-17
Have a special indicator for official APT in balance change tab


ying-w:
As we see more spam coins that share the APT symbol, can we have a special indicator on balance change tab for when real APT (both coin and FA) is being transferred.

Ex. https://explorer.aptoslabs.com/txn/1807783457/balanceChange?network=mainnet

PR #827, opened 2024-10-17
[coin balance changes] Add coin balance change verification

gregnazario:
To differentiate between known coins and unknown coins, it will now pull from the Panora coin list to give more information.

Note that, right now, this is a beta feature... There are FAs that are actually migrated coins that need to be updated in the feature.

Partially resolves https://github.com/aptos-labs/explorer/issues/826

Follow-ups:
1. Coin details page
2. FA details page
3. Coin balances on accounts page


ying-w (approved):
Checked examples of coins (v1), fa (v2) and some scam coins and they all look good

Issue #825, opened 2024-10-17
Issues with gas being shown on explorer


ying-w:
Following up on https://github.com/aptos-labs/explorer/pull/824

Example transaction:

- https://explorer.aptoslabs.com/txn/376988406/userTxnOverview?network=mainnet
- https://fullnode.mainnet.aptoslabs.com/v1/transactions/by_version/376988406

What I expect to see is:

- 114,000 octa (or 0.00114 APT) burned for gas
- (same for net gas since no storage refund)
- 76 gas units used (since 76 gas units * 1500 gas unit price = 114,000 octa)
- 1500 octa for gas unit price (or 0.000015 APT)
- 15000 gas units for max gas

I think a few confusing things are gas units (max_gas, gas used) vs octa or apt (amount burned, gas unit cost)

For APT vs octa, I think most explorers show in native so ok with showing in APT but comparing the number of 0s for decimals does make for kinda poor UX (would it be possible to add spaces in somewhere? too bad we don't do commas for decimals)

As an FYI, to get to that 76 gas unit, it is approx the sum of 

- 8 execution units
- 2 io units
- 100000 octa / 1500 = 67 storage units

This information can be found in the FeeStatement event

PR #828, opened 2024-10-17
[gas] Fix gas units showing

gregnazario:
resolves https://github.com/aptos-labs/explorer/issues/825


ying-w (approved):
Looks good, adding octa as grey i think is a nice touch

PR #829, opened 2024-10-18
[account] Show if the authentication key is rotated

gregnazario:
An example is here:
https://deploy-preview-829--aptos-explorer.netlify.app/account/0xfcca5a63a68598ae6d56324df47ebb03c375e8ad82a9d7a1bf6520f16e5377a6/info?network=mainnet


bv-aptos (approved):
Not sure how much this will help an end user, but should help dev identify easily if/when an account has been rotated...I assume that's the point?

gregnazario:
Yeah, from a support side it'll be a little more helpful.

PR #834, opened 2024-10-21
Add verification to balances page

gregnazario:
Adds the same thing to account coin balances as it did to balance changes

### DIFF
--- a/src/components/Table/VerifiedCell.tsx
+++ b/src/components/Table/VerifiedCell.tsx
@@ -2,10 +2,12 @@ import GeneralTableCell from "./GeneralTableCell";
 import {Stack} from "@mui/material";
 import StyledTooltip from "../StyledTooltip";
 import {
+  Dangerous,
   DangerousOutlined,
-  VerifiedUserOutlined,
+  Verified,
+  VerifiedUser,
+  Warning,
   WarningAmberOutlined,
-  WarningOutlined,
 } from "@mui/icons-material";
 import VerifiedOutlined from "@mui/icons-material/VerifiedOutlined";
 import * as React from "react";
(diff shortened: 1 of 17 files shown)


src/components/Table/VerifiedCell.tsx
```diff
@@ -0,0 +1,93 @@
...
+            <WarningAmberOutlined fontSize="small" />
+          </StyledTooltip>
+        </Stack>
+      ) : (
+        <Stack direction="row" spacing={1} alignItems="center">
+          <StyledTooltip title="This asset is not verified, it may or may not be recognized by the community.">
+            <WarningOutlined fontSize="small" />
+          </StyledTooltip>
+        </Stack>
+      )}
```
kaw2k, line 90: Mind extracting this into a `switch` or something similar. Deeply nested ternaries in render statements are an eye-full.

src/pages/Account/Components/CoinsTable.tsx
```diff
@@ -64,62 +61,79 @@ function CoinTypeCell({
...
           <GeneralTableHeaderCell header="Name" />
+          <GeneralTableHeaderCell header="Asset Type" />
+          <GeneralTableHeaderCell header="Asset" />
+          <GeneralTableHeaderCell
+            header="Verified"
+            tooltip={
+              <LearnMoreTooltip
+                text="This uses the Panora token list to verify authenticity of known assets on-chain.  It does not guarantee anything else about the asset and is not financial advice."
+                link="https://github.com/PanoraExchange/Aptos-Tokens"
+              />
```
kaw2k, line 105: If you wouldn't mind modifying `LearnMoreTooltip` to add `target="_blank" to the `Link`. That way we open these external links in new tabs.

kaw2k:
Thanks for adding this!

src/components/Table/VerifiedCell.tsx
gregnazario: Yeah, it truly is, I'm not the greatest at this stuff, let me see what I can do

kaw2k (approved):

PR #835, opened 2024-10-22
Verified names

gregnazario:
This provides the ability for verified names, separate of ANS.  ANS name lookups are skipped, and it provides a verification check icon to show that it's a verified name.

For now, I've only provided 0xA as a tester.

An example here: https://deploy-preview-835--aptos-explorer.netlify.app/account/0x000000000000000000000000000000000000000000000000000000000000000a?network=mainnet

### DIFF
--- a/src/api/hooks/useGetANS.ts
+++ b/src/api/hooks/useGetANS.ts
@@ -5,10 +5,19 @@ import {
   fetchJsonResponse,
   getLocalStorageWithExpiry,
   setLocalStorageWithExpiry,
+  standardizeAddress,
 } from "../../utils";
 import {ResponseError} from "../client";
 
 const TTL = 60000; // 1 minute
+
+// This is an override of ANS names, in case we want to display a verified name for an address
+// TODO: this probably belongs somewhere else... but, for now, it's here
+const knownAddresses: Record<string, string> = {
+  "0xa": "AptosCoin",
+  "0x000000000000000000000000000000000000000000000000000000000000000a":
+    "AptosCoin",
+};
 
 function getFetchNameUrl(
   network: NetworkName,
@@ -33,11 +42,22 @@ export function useGetNameFromAddress(
   const queryResult = useQuery<string | null, ResponseError>({
     queryKey: ["ANSName", address, shouldCache, state.network_name],
     queryFn: () => {
+      const standardizedAddress = standardizeAddress(address);
+      const knownName = knownAddresses[standardizedAddress.toLowerCase()];
+      if (knownName) {
+        return knownName;
+      }
       const cachedName = getLocalStorageWithExpiry(address);
       if (cachedName) {
         return cachedName;
       }
-      return genANSName(address, shouldCache, state.network_name, isValidator);
+      // Ensure there's always .apt at the end
+      return genANSName(
+        address,
+        shouldCache,
+        state.network_name,
+        isValidator,
+      ).then((name) => (name ? `${name}.apt` : null));
     },
   });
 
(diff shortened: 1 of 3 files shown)


src/components/TitleHashButton.tsx
```diff
@@ -135,13 +136,24 @@ function Name({address, isValidator}: {address: string; isValidator: boolean}) {
...
         }}
       >
-        <Link
-          href={`https://www.aptosnames.com/name/${name}`}
-          target="_blank"
-          underline="none"
-        >
-          {`${name}.apt`}
-        </Link>
+        {name.endsWith(".apt") ? (
```
kaw2k, line 139: I _think_ we have special treatment for `.petra` names as well. To be safe, change this to:
```ts
const isAns = name.endsWith(".apt") || name.endsWith(".petra")
{isAns ? () : ()}
```

src/components/TitleHashButton.tsx
```diff
@@ -135,13 +136,24 @@ function Name({address, isValidator}: {address: string; isValidator: boolean}) {
...
+          <Link
+            href={`https://www.aptosnames.com/name/${name}`}
+            target="_blank"
+            underline="none"
+          >
+            {name}
+          </Link>
+        ) : (
+          <>
```
kaw2k, line 148: You don't need the fragment tag `<></>` If you only have a single child.

src/components/TitleHashButton.tsx
```diff
@@ -135,13 +136,24 @@ function Name({address, isValidator}: {address: string; isValidator: boolean}) {
...
+            <Typography>
```
kaw2k, line 149: When working with things in a line, `flex` is your friend. Relying on white space can cause issues like bellow (based on the size of the other elements in the container).

image

Instead, try the following:

```tsx
  <Typography sx={{display: 'flex', alignItems: 'row', gap: 4}}>
    <span>{name}</span>
    <Tooltip title={'This is a verified address label.'}>
      <VerifiedOutlined fontSize="small" />
    </Tooltip>
  </Typography>
```

Wrapping the `name` in a span gives `flex` a node to latch on to. `alignItems` aligns on the secondary axis (in this case the y axis) which centers everything. If that still doesn't line up nicely, you have to modify the heights of the children of the flex container.

src/api/hooks/useGetANS.ts
```diff
@@ -33,11 +42,22 @@ export function useGetNameFromAddress(
...
         return cachedName;
       }
-      return genANSName(address, shouldCache, state.network_name, isValidator);
+      // Ensure there's always .apt at the end
+      return genANSName(
+        address,
+        shouldCache,
+        state.network_name,
+        isValidator,
+      ).then((name) => (name ? `${name}.apt` : null));
```
kaw2k, line 60: Make sure to test this with a subdomain.

src/components/TitleHashButton.tsx
```diff
@@ -135,13 +137,22 @@ function Name({address, isValidator}: {address: string; isValidator: boolean}) {
...
+            target="_blank"
+            underline="none"
+          >
+            {name}
+          </Link>
+        ) : (
+          <Typography sx={{display: "flex", alignItems: "row", gap: 1}}>
+            <span>{name}</span>
+            <Tooltip title={"This is a verified address label."}>
+              <VerifiedOutlined fontSize="small" />
```
kaw2k, line 152: Sorry to nitpick, the alignment was still slightly off. I think perhaps the line height of the font is causing it. This should address it. 

```suggestion
              <VerifiedOutlined fontSize="small" sx={{position: 'relative', top: 2}} />
```

kaw2k (approved):

PR #840, opened 2024-10-23
[counterparty] Fix counterparty

gregnazario:
1. Remove contract address as counterparty, is confusing
2. Add FA support and object transfer support
3. Support all basic coin transfers

Screenshot 2024-10-23 at 4 28 49 PM
Screenshot 2024-10-23 at 4 29 06 PM
Screenshot 2024-10-23 at 4 29 25 PM


kaw2k (approved):

ying-w:
i _kinda_ like the old information dense way better to see which contract gets interacted with, it is nice that we're parsing more core contracts

> 0x3… transfers are impossible because of claim and offer

PR #843, opened 2024-10-24
Add proper support for search and FA data

gregnazario:
This is specifically to add support for future FA, and details around it.

Will add screenshots in a moment.

Does the UI look a little messy?  Yes, but I don't want to spend too much time cleaning up things.  If we want better looks, we need a real frontend engineer :)

Some examples:
1. TeviStar, a non-panora listed asset https://deploy-preview-843--aptos-explorer.netlify.app/fungible_asset/0x43782fca70e1416fc0c75954942dadd4af8d305a608b6153397ad5801b71e72d?network=mainnet
2. amAPT, a panora listed asset https://deploy-preview-843--aptos-explorer.netlify.app/coin/0x111ae3e5bc816a5e63c2da97d0aa3886519e0cd5e4b046659fa35796bd11542a::amapt_token::AmnisApt?network=mainnet
3. Cellana, a panora listed fungible asset https://deploy-preview-843--aptos-explorer.netlify.app/fungible_asset/0x2ebb2ccac5e027a87fa0e2e5f656a3a4238d6a48d93ec9b610d570fc0aa0df12?network=mainnet
4. A scam coin: https://deploy-preview-843--aptos-explorer.netlify.app/coin/0x397071c01929cc6672a17f130bd62b1bce224309029837ce4f18214cc83ce2a7::USDC::USDC?network=mainnet
5. USDT search, you can search by symbol or by name!
Screenshot 2024-10-24 at 12 42 10 AM
6. Layer Zero USDT, a bridged asset: https://deploy-preview-843--aptos-explorer.netlify.app/coin/0xf22bede237a07e121b56d91a491eb7bcdfd1f5907926a9e58338f964a01b17fa::asset::USDT?network=mainnet


src/ExplorerRoutes.tsx
```diff
@@ -35,6 +36,10 @@ export default function ExplorerRoutes() {
           <Route path=":struct" element={<CoinPage />} />
           <Route path=":struct/:tab" element={<CoinPage />} />
         </Route>
+        <Route path="/fungible_asset">
+          <Route path=":address" element={<FAPage />} />
+          <Route path=":address/:tab" element={<FAPage />} />
+        </Route>
```
0xbe1, line 42: out of scope but do you think we may add a assets page to explorer? reference https://aptoscan.com/coins

gregnazario: That's the last thing I may add here, though I'd have to check on the efficiency of the indexer queries

src/pages/FungibleAsset/Tabs/InfoTab.tsx
```diff
@@ -1,47 +1,99 @@
...
+              >
+                <span>{data?.metadata?.name}</span>
+                <VerifiedCoinCell
+                  data={{
+                    id: address,
+                    known: !!coinData,
+                    isBanned: coinData?.isBanned,
+                    isInPanoraTokenList: coinData?.isInPanoraTokenList,
+                  }}
+                />
```
kaw2k, line 58: `VerifiedCoinCell` is rendering a `td` under the hood which is invalid markup as it is outside of a `table` tag. In `src/components/Table/VerifiedCell.tsx` lets add a component `VerifiedAsset` (or something) which accepts the same props `VerifiedCellProps` but only renders the `StyledTooltip` that the current component returns. Export that, and we can use that new component in the existing `VerifiedCoinCell` component. That will fix the invalid markup and the odd spacing (as `GeneralTableCell` is applying padding which throws off alignment)

src/pages/Coin/Tabs/InfoTab.tsx
```diff
@@ -30,18 +37,64 @@ export default function InfoTab({struct, data, supply}: InfoTabProps) {
...
+              >
+                <span>{data?.data?.name}</span>
+                <VerifiedCoinCell
+                  data={{
+                    id: struct,
+                    known: !!coinData,
+                    isBanned: coinData?.isBanned,
+                    isInPanoraTokenList: coinData?.isInPanoraTokenList,
+                  }}
+                />
```
kaw2k, line 61: Use that same component `Verified` (whatever) that we create here as well to avoid the `td`.

src/api/hooks/useGetFaMetadata.ts
```diff
@@ -0,0 +1,34 @@
...
+      const mappedData = data as [FaMetadata];
+      const val = mappedData[0];
+      if (val !== undefined && val !== null) {
+        setMetadata(val);
+      }
+    }
+  }, [data, state]);
+
+  return metadata;
+}
```
kaw2k, line 34: nit: You don't need `useEffect` here. both `useGlobalState` and `useViewFunction` are optimized, so all `useEffect` is giving us is extra code. You can simplify to the following: 


```suggestion
import {useViewFunction} from "./useViewFunction";

type FaMetadata = {
  name: string;
  symbol: string;
  decimals: number;
  icon_uri: string;
  project_uri: string;
};

export function useGetFaMetadata(address: string): FaMetadata | null {
  const {data} = useViewFunction(
    "0x1::fungible_asset::metadata",
    ["0x1::object::ObjectCore"],
    [address],
  );

  if (data) {
    const [val] = data as [FaMetadata];
    if (val !== undefined && val !== null) {
      return val;
    }
  }

  return null;
}
```

src/api/hooks/useGetFaSupply.ts
```diff
@@ -0,0 +1,26 @@
...
+      const mappedData = data as [{vec: [string]}];
+      const val = mappedData[0]?.vec[0];
+      if (val !== undefined && val !== null) {
+        setSupply(BigInt(val));
+      }
+    }
+  }, [data, state]);
+
+  return supply;
+}
```
kaw2k, line 26: Same deal here, we can simplify this by removing the use of `useEffect` and `useState`

```suggestion
import {useViewFunction} from "./useViewFunction";

export function useGetFASupply(address: string): bigint | null {
  const {data} = useViewFunction(
    "0x1::fungible_asset::supply",
    ["0x1::object::ObjectCore"],
    [address],
  );

  if (data !== undefined) {
    const mappedData = data as [{vec: [string]}];
    const val = mappedData[0]?.vec[0];
    if (val !== undefined && val !== null) {
      return BigInt(val);
    }
  }

  return null;
}
```

src/ExplorerRoutes.tsx
gregnazario: @0xbe1 please open a GitHub issue for the coin page.

src/components/Table/VerifiedCell.tsx
```diff
@@ -49,6 +49,10 @@ const labsBannedTokens: Record<string, string> = {
     "scam",
   "0x48327a479bf5c5d2e36d5e9846362cff2d99e0e27ff92859fc247893fded3fbd::APTOS::APTOS":
     "scam",
+  "0xbc106d0fef7e5ce159423a1a9312e011bca7fb57f961146a2f88003a779b25c2::QUEST::QUEST":
+    "scam",
+  "0xbe5e8fa9dd45e010cadba1992409a0fc488ca81f386d636ba38d12641ef91136::maincoin::Aptmeme":
+    "scam",
```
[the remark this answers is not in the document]
gregnazario: Yeah, I think that's totally the goal.  But, for now this is a hardcoded list.

Ideally all labs products should use the same, dynamic scam list.

src/utils.ts
```diff
@@ -251,3 +252,20 @@ export function getStableID(): string {
 export function sortPetraFirst(a: AnyAptosWallet) {
   return a.name === "Petra" ? -1 : 1;
 }
+
+export function getAssetSymbol(
+  panoraSymbol: string | null | undefined,
+  bridge: string | null | undefined,
+  symbol: string | null | undefined,
+) {
```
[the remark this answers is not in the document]
gregnazario: Yeah, I think this and a ton of other things around these types need to be cleaned up in a future PR.  It's been a bit of a mess between the different types, and I'm shimming them together.

kaw2k (approved):
Great work!

PR #847, opened 2024-10-24
[user txns] Hide type on user transactions table

gregnazario:
This symbol means nothing, because every transaction by definition is a user transaction.

This gives us a little more real estate.

Applies to the main page https://deploy-preview-847--aptos-explorer.netlify.app/?network=mainnet

And for users https://deploy-preview-847--aptos-explorer.netlify.app/account/0xc67545d6f3d36ed01efc9b28cbfd0c1ae326d5d262dd077a29539bcee0edce9e?network=mainnet


ying-w:
I actually like the type on that page, when i'm looking at addresses that became validators i can quickly separate what is user transaction

PR #851, opened 2024-10-24
[search] Add exact search for known names

gregnazario:
I started with fuzzy matching, but it gave too many results, so it is only exact match.

I also removed some confusing labels, 0xa was mostly a placeholder.


ying-w (approved):
kinda feel like could be some unforeseen consequences here but stamping for now

PR #854, opened 2024-10-25
[CoinsTable] allow filter by verified

0xbe1:
## difference

|before|after|
|--|--|
|image|image|

## background

currently, coins page is filled with spam coins, hard to see what what coins matter even with the verified logo.
 
## solution

add "verified | all" tabs and default to "verified" so that only verified coins are displayed by default. only show the rest unless we really want to and click "all"

## to reviewers

the UI is pretty much copied from CollapsibleCards.tsx, only modifying text "expand all | collapse all" to "verified | all"

### DIFF
--- a/src/pages/Account/Components/CoinsTable.tsx
+++ b/src/pages/Account/Components/CoinsTable.tsx
@@ -10,7 +10,7 @@ import {
 import GeneralTableRow from "../../../components/Table/GeneralTableRow";
 import GeneralTableHeaderCell from "../../../components/Table/GeneralTableHeaderCell";
 import HashButton, {HashType} from "../../../components/HashButton";
-import {grey} from "../../../themes/colors/aptosColorPalette";
+import {grey, primary} from "../../../themes/colors/aptosColorPalette";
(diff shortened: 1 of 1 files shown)


src/pages/Account/Components/CoinsTable.tsx
```diff
@@ -89,57 +96,116 @@ export type CoinDescriptionPlusAmount = {
...
 
 export function CoinsTable({coins}: {coins: CoinDescriptionPlusAmount[]}) {
+  const [showVerifiedOnly, setShowVerifiedOnly] = React.useState(true);
+
+  const filteredCoins = showVerifiedOnly
+    ? coins.filter((coin) => coin.isInPanoraTokenList)
+    : coins;
+
+  const heavyTextColor = grey[450];
+  const lightTextColor = grey[400];
```
gregnazario, line 106: These need to be more significant of a difference.

It's hard to tell which one's selected and which one isn't.

I think we should probably also add recognized as a selection (basically known + not banned).

Let me know if you want to go down this path or need help to do it

gregnazario (approved):

Issue #793, opened 2024-09-25
'No Data Found' returned for accounts with a lot of transactions


ying-w:
Ex1. `0x820931c36d84f8f5b44d4b37701c00b29e55f714eec2165c47508bf57d18a42c` https://explorer.aptoslabs.com/account/0x820931c36d84f8f5b44d4b37701c00b29e55f714eec2165c47508bf57d18a42c?network=mainnet vs https://aptoscan.com/account/0x820931c36d84f8f5b44d4b37701c00b29e55f714eec2165c47508bf57d18a42c

Ex2. `0x2d91309b5b07a8be428ccd75d0443e81542ffcd059d0ab380cefc552229b1a`
https://explorer.aptoslabs.com/account/0x2d91309b5b07a8be428ccd75d0443e81542ffcd059d0ab380cefc552229b1a?network=mainnet vs https://aptoscan.com/account/0x2d91309b5b07a8be428ccd75d0443e81542ffcd059d0ab380cefc552229b1a

I think showing some loading screen or pagination would be an idea solution here

gregnazario:
Yeah, in this case it's using this query to determine the number of pages, and it's taking too long

PR #857, opened 2024-10-25
[account] Show up to 10k transactions

gregnazario:
The aggregate query sometimes timesout for accounts, because there are too many transactions.

Resolves: https://github.com/aptos-labs/explorer/issues/793


ying-w (approved):
10k limit is only if query times out

PR #876, opened 2024-10-28
[search] Cleanup usage of AccountAddress.from to handle errors add search by coin struct

gregnazario:
This should get rid of any random crashes due to an account address not being formatted correctly.


src/api/hooks/useGetANS.ts
```diff
@@ -22,8 +22,6 @@ export const knownAddresses: Record<string, string> = {
     "Legacy Token (0x3)",
   "0x0000000000000000000000000000000000000000000000000000000000000004":
     "Digital Assets (0x4)",
-  "0x000000000000000000000000000000000000000000000000000000000000000a":
```
ying-w, line 25: why the removal here

gregnazario: I think it showed up funny in search.  I could go either way here

ying-w (approved):
hope this catches all of the cases where address is used

PR #887, opened 2024-11-01
[analytics] Add fullnode count

gregnazario:
### Description

Adds fullnode approx count to the page
Screenshot 2024-11-01 at 6 10 08 PM

### Checklist

- [ ] Have you run `pnpm fmt`?


ying-w:
Do you want to put it with previous and instead of callling it 'active validators' put 'Full nodes' and then "validators" and "total" with tooltip saying to total is a lower bound approx

bv-aptos (approved):
Nice! Cranking it out day and night!!

Issue #865, opened 2024-10-26
[Feature Request] Add "actions" to txn page


0xbe1:
in aptoscan, a txn is interpreted as actions against major protocols, which makes the txn very easy to understand, for example reference: https://aptoscan.com/transaction/1843272960

<img width="1304" alt="image" src="https://github.com/user-attachments/assets/e9e7a7c6-3077-4ad8-aa70-ef56d632379a">

would be great if the explorer also support this and each protocol can chime in and interpret their actions

0xbe1:
closed by #908

PR #908, opened 2024-11-12
[EventsTab] add txn actions and use thala as an example

0xbe1:
### Description

set up a template to add event interpretation (aka "action") to txn events tab, and add thalaswap actions as the first ones.

see https://deploy-preview-908--aptos-explorer.netlify.app/txn/1916163127?network=mainnet

image

### Related Links

#865 

### Checklist

- [x] Have you run `pnpm fmt`?

### DIFF
--- a/src/pages/Transaction/Tabs/UserTransactionOverviewTab.tsx
+++ b/src/pages/Transaction/Tabs/UserTransactionOverviewTab.tsx
@@ -17,6 +17,9 @@ import JsonViewCard from "../../../components/IndividualPageContent/JsonViewCard
 import {parseExpirationTimestamp} from "../../utils";
 import {TransactionActions} from "./Components/TransactionActions";
 import {grey} from "../../../themes/colors/aptosColorPalette";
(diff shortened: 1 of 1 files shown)


gregnazario (commented):
My only thought here, maybe we pull these out of the events and onto the main page to be more helpful to the average user.

In this case, you can see here someone once upon a time, thought about this for NFT transfers / mints.  Maybe we can move it there?
https://explorer.aptoslabs.com/txn/1811127385?network=mainnet

0xbe1:
> My only thought here, maybe we pull these out of the events and onto the main page to be more helpful to the average user.
> 
> In this case, you can see here someone once upon a time, thought about this for NFT transfers / mints. Maybe we can move it there? https://explorer.aptoslabs.com/txn/1811127385?network=mainnet

are you ok with having "action" in both main page and events page?

0xbe1:
@gregnazario understood, moved it under txn page

gregnazario (approved):
Let's get everyone to add parsers!

Maybe we should write instructions

Issue #906, opened 2024-11-07
[RFC] Coin / FA Page Updates and Features


gregnazario:
Please comment below about some of the features requested for the coin or FA details pages.

We'll collect them here, and prioritize / determine what people want for focusing here.

gregnazario:
So, FDV isn't necessarily shown here, those are based on tokenomics.

Holders, and associated transactions mostly work.

Coins link to FA and vice versa.

Remaining items:

- [ ] Better performance of transactions tab
- [ ] Mint and Burn information
- [ ] If Mint and Burn capabilities were created
- [ ] Initial creation transaction

PR #918, opened 2024-11-15
Fixup coin holders page / Add transactinos

gregnazario:
### Description
Added bare minimum associated transactions.  To combine data, we'd need to be able to query multiple transactions 

Follows up on an item from https://github.com/aptos-labs/explorer/issues/906


ying-w (approved):

ying-w:
Transaction tab: need to click on version twice

Only showing 9 txns (i think holders might also be missing)

PR #925, opened 2024-11-19
[titles] Add updating titles to each pages

gregnazario:
### Description
Adds titles to each page, I'm not sure I really know what I'm doing here with all the use effects, but it does seem to work decently.

I observed google searches don't give good title names, so adding this will help the google web crawler to give better names

### DIFF
--- a/src/components/hooks/usePageMetadata.ts
+++ b/src/components/hooks/usePageMetadata.ts
@@ -1,0 +1,13 @@
+import {useEffect} from "react";
+
+interface PageMetadataArgs {
+  title?: string;
+}
+
+export function usePageMetadata(args: PageMetadataArgs = {}): void {
+  useEffect(() => {
+    document.title = args.title
+      ? `${args.title} | Aptos Explorer`
+      : "Aptos Explorer";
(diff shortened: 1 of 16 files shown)


kaw2k:
Great work! And `useEffect` is an appropriate use case here. An alternative (not saying to do it) is to use something like [react-helmate](https://github.com/nfl/react-helmet). I think we should stick with your approach, but abstract it just a touch. I would recommend a hook like the following:

```tsx
import { useEffect } from 'react'

interface PageMetadataArgs {
  title?: string
}

export function usePageMetadata(args: PageMetadataArgs = {}): void {
  useEffect(() => {
    document.title = args.title ? `${args.title} | Aptos Explorer` : 'Aptos Explorer'
  }, [args]);
};
```
Note, page titles typically go from most specific to least specific. You could change `title` from `string` to `string | string[]` and chain them if you wanted, but typically you want titles under 60 characters for proper rendering in tabs. 

This would also let us attach other metadata to the document in the future, twitter tags, facebook tags, etc.

kaw2k (approved):

Issue #867, opened 2024-10-26
[Feature Request] Display coin value besides amount


0xbe1:
Price info is very important for coins. Believe we may work in the following order:
1. In /account page, render coin value in the CoinsTable.
2. In /account page, render account value besides APT amount.
3. In /coin and /fungible_asset page, given asset supply, render asset price and market cap.

We may leverage [Panora API](https://docs.panora.exchange/developer/token-prices) for price fetching.

**Note** this is a mainnet-only feature

PR #948, opened 2024-11-30
issue #867 2. added usd value using coingecko api

sahil7886:
### Description

Implements 2. in Issue #867 
Added functionality to display the USD value of the connected wallet's APT balance. Created a new hook, useGetAPTPrice.ts that uses the coingecko api to fetch the current price of APT in USD. The BalanceCard.tsx then uses that hook to fetch the current price, do a few calculations on it and displays it. The USD Value only shows if the user has selected the mainnet. 

### Related Links

https://github.com/aptos-labs/explorer/issues/867

### DIFF
--- a/src/api/hooks/useGetAPTPrice.ts
+++ b/src/api/hooks/useGetAPTPrice.ts
@@ -1,0 +1,30 @@
+const COINGECKO_API_ENDPOINT = "https://api.coingecko.com/api/v3/simple/price";
+
+export async function getAPTPrice(): Promise<number> {
+  const query = {
+    ids: "aptos", // CoinGecko ID for Aptos
+    vs_currencies: "usd", // Target currency
+  };
+
+  const queryString = new URLSearchParams(query);
+  const url = `${COINGECKO_API_ENDPOINT}?${queryString}`;
+
+  try {
+    const response = await fetch(url, {
+      method: "GET",
+    });
+
+    // Ensure the response is successful
+    if (!response.ok) {
+      throw new Error(`HTTP error! Status: ${response.status}`);
+    }
+
+    const data = await response.json();
+
+    // Extract and return the USD price
+    return Number(data.aptos.usd);
+  } catch (error) {
+    console.error("Error fetching APT price from CoinGecko:", error);
+    return 0; // Default fallback price
+  }
+}
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: "6.0"
+lockfileVersion: "9.0"
 
 settings:
   autoInstallPeers: true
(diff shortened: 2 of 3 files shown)


src/api/hooks/useGetAPTPrice.ts
```diff
@@ -0,0 +1,30 @@
+const COINGECKO_API_ENDPOINT = "https://api.coingecko.com/api/v3/simple/price";
+
+export async function getAPTPrice(): Promise<number> {
+  const query = {
+    ids: "aptos", // CoinGecko ID for Aptos
+    vs_currencies: "usd", // Target currency
+  };
```
gregnazario, line 7: Awesome, we should make this `getPrice` and take in any coingecko Id.

Then we can use it for anything that has a coingecko ID

src/api/hooks/useGetAPTPrice.ts
```diff
@@ -0,0 +1,30 @@
...
+      throw new Error(`HTTP error! Status: ${response.status}`);
+    }
+
+    const data = await response.json();
+
+    // Extract and return the USD price
+    return Number(data.aptos.usd);
+  } catch (error) {
+    console.error("Error fetching APT price from CoinGecko:", error);
+    return 0; // Default fallback price
```
gregnazario, line 28: Can we make the fallback `null`, so then we can handle elsewhere to hide the dollar amount if it doesn't apply.

gregnazario:
Let me know if don't feel like changing it, and I will follow up later about it

src/api/hooks/useGetAPTPrice.ts
```diff
@@ -0,0 +1,30 @@
+const COINGECKO_API_ENDPOINT = "https://api.coingecko.com/api/v3/simple/price";
+
+export async function getAPTPrice(): Promise<number> {
```
kaw2k, line 3: Thank you for contributing! Some food for thought!

This lives in the `hooks` folder, however this as implemented is not a hook. I like the idea you were going with of having a separate fetch function and hook (ie: `getPrice` and `useGetPrice`). Perhaps export both functions, and the caller can choose which function to use. 

Bellow is a simple example of how you might approach this. Note, this repo uses `react-query` which helps manage caching and state. Your code in `BalanceCard.tsx` can be simplified o remove the use of `useState` if you use `react-query` 

```tsx
import { useQuery } from "@tanstack/react-query";

export function useGetPrice(id: string) {
  return useQuery({
    queryKey: ["price", id],
    queryFn: () => getAPTPrice(id),
  });
}

// No need for state!
const {data: price} = useGetPrice('coin_id')
```

src/api/hooks/useGetAPTPrice.ts
sahil7886: That makes a lot of sense. Changed that and set the default value of the argument to the aptos id, so as to not change existing implementations of the function.

src/api/hooks/useGetAPTPrice.ts
sahil7886: Understood. I've implemented the useGetPrice() hook using react query. Let me know if I've done it right and if any more changes need to made. I've also kept the getPrice() function there as it is. Would it make sense to add the getPrice() function to api/index.ts and keep just the hook in the useGetPrice.ts file?

gregnazario (approved):
Love it!  Thank you.  @kaw2k give it a second look

kaw2k (approved):
Nice work!

PR #950, opened 2024-12-01
[supply] Verify uptos supply

gregnazario:
### Description

Verifies uptos's supply.  No mints had occurred after initial mint.


ying-w (approved):
have we asked indexer team for endpoint for these edge cases?

PR #963, opened 2024-12-18
[search/coin] Fix LP coin lookup and linking

gregnazario:
### Description

Fixes LP coin lookup, and more complex coin names


0xbe1 (approved):
lgtm

PR #965, opened 2024-12-20
[txns] Fix navigation in transactions with back

rs-8:
Hello!

Fix back navigation in transactions page

For the first navigation to transactions, a replacement is used so that you can exit with one click

Navigations between user and all transactions work as usual

**Before:**

https://github.com/user-attachments/assets/249dcf58-1e16-4a82-840b-f9003621f296

**After**

https://github.com/user-attachments/assets/bd34345c-d35e-4aa4-81da-a34e0cba64b7

**Thanks!**

### DIFF
--- a/src/pages/Transactions/Index.tsx
+++ b/src/pages/Transactions/Index.tsx
@@ -1,25 +1,34 @@
-import {useEffect, useState} from "react";
 import {useSearchParams} from "react-router-dom";
 import {Box, Button, Stack, Typography} from "@mui/material";
 import PageHeader from "../layout/PageHeader";
 import AllTransactions from "./AllTransactions";
 import UserTransactions from "./UserTransactions";
 import {useGetIsGraphqlClientSupported} from "../../api/hooks/useGraphqlClient";
-import {useGlobalState} from "../../global-config/GlobalConfig";
 import {usePageMetadata} from "../../components/hooks/usePageMetadata";
+import {useEffect} from "react";
 
 export default function TransactionsPage() {
-  const [state] = useGlobalState();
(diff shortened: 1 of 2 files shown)


src/pages/Transactions/Index.tsx
```diff
@@ -1,25 +1,34 @@
...
+    if (isGraphqlClientSupported && !searchParams.get("type")) {
+      searchParams.set("type", "user");
+      return setSearchParams(searchParams, {replace: true});
+    }
+
+    if (!isGraphqlClientSupported && !searchParams.get("type")) {
+      searchParams.set("type", "all");
+      return setSearchParams(searchParams, {replace: true});
+    }
+  }, [isGraphqlClientSupported, searchParams, setSearchParams, allTxnOnly]);
```
kaw2k, line 28: Just a touch of cleanup here:

* Looks like we no longer need `allTxnOnly` so we can omit it from the dependency array
* You should only return a [cleanup function](https://react.dev/reference/react/useEffect#parameters) from `useEffect`, otherwise, avoid returns. 

I suggest the following:

```tsx
useEffect(() => {
  if (!searchParams.get("type")) {
    searchParams.set("type", isGraphqlClientSupported ? "user" : "all");
    setSearchParams(searchParams, { replace: true });
  }
}, [isGraphqlClientSupported, searchParams, setSearchParams]);
```

kaw2k (changes requested):
Thank you for the before and after, and for contributing!

src/pages/Transactions/Index.tsx
rs-8: Looks correct and simpler, added changes to pull request
Thank you for review

kaw2k (approved):
Excelent, thank you!

PR #1042, opened 2025-02-24
[validators] Fix pagination of delegated validators list

gregnazario:
### Description

Adds proper pagination, and fetching more to the list.

We will need to add pagination on the table at some point, but for now it properly pulls all the delegation validators.

Will need to follow up on other pages


teenu4:
Thanks for such a quick fix, @gregnazario !

PR #1049, opened 2025-03-06
Fix transaction submission against local testnet

0xmaayan:
### Description

When users are using a wallet (Petra) set to localnet and set Explorer to localnet, they are having `Wallet and Explorer should use the same network to submit a transaction` error

The wallet adapter uses "local" as the network name, so no need to add the conversion to "localhost"

### DIFF
--- a/src/api/hooks/useSubmitTransaction.ts
+++ b/src/api/hooks/useSubmitTransaction.ts
@@ -40,8 +40,7 @@ const useSubmitTransaction = () => {
 
   async function submitTransaction(transaction: InputTransactionData) {
     if (
-      network?.name.toLocaleLowerCase() !==
-        (state.network_name === "local" ? "localhost" : state.network_name) &&
+      network?.name.toLocaleLowerCase() !== state.network_name &&
       // TODO: This is a hack to get around network being cached
       wallet?.name !== "Google (AptosConnect)"
     ) {


src/api/hooks/useSubmitTransaction.ts
```diff
@@ -43,7 +43,7 @@ const useSubmitTransaction = () => {
       network?.name.toLocaleLowerCase() !==
         (state.network_name === "local" ? "localhost" : state.network_name) &&
       // TODO: This is a hack to get around network being cached
-      wallet?.name !== "Google (AptosConnect)"
+      wallet?.name === "Google (AptosConnect)"
```
kent-white, line 46: So the reason for this hack is the Aptos Connect SDK handles network differently and there isnt an easy way to refresh the plugin. Thus in AptosConnect we are ignoring the network name check

Your change would make it so that explorer ignores the network name check for everything but Apto Connect. Which I think is not what you want?

src/api/hooks/useSubmitTransaction.ts
```diff
@@ -40,8 +40,7 @@ const useSubmitTransaction = () => {
 
   async function submitTransaction(transaction: InputTransactionData) {
     if (
-      network?.name.toLocaleLowerCase() !==
-        (state.network_name === "local" ? "localhost" : state.network_name) &&
+      network?.name.toLocaleLowerCase() !== state.network_name &&
```
kent-white, line 43: the network name Petra responds with is Localhost though :(

0xmaayan: From my tests it is `local`, and it also matches to the adapter NetworkInfo type that Explorer is using
```
export type NetworkInfo = {
  name: Network;
  chainId?: string;
  url?: string;
};
```

kent-white: How did you test?

Screenshot 2025-03-06 at 12 00 15 PM

kent-white (approved):
not sure how you are testing this, maybe walletadapter changes it to local?

src/api/hooks/useSubmitTransaction.ts
0xmaayan: Through Explorer, it uses the wallet adapter that standardize the network name - `local`.

PR #1051, opened 2025-03-11
improve delegation page

jillxuu:
### Description
1. most of the fields sortable
2. improve the UX, UI
3. improve query efficiency to load way less time using batch
4. in interim, can only load the new delegation page using `showEnhanced=true` params in the url

Screenshot 2025-03-11 at 2 43 23 PM


gregnazario (approved):
Seems to work!

PR #1069, opened 2025-03-18
[wallets] Remove legacy wallets

gregnazario:
### Description
Removes all plugin wallets, and updates wallet adapter

Do not push until 5/1/2025.


kaw2k (approved):
Looks good!

Issue #1135, opened 2025-05-19
[Bug] Truncated addresses not AIP-40 compliant


banool:
The truncated account addresses displayed here aren't AIP-40 compliant: https://explorer.aptoslabs.com/txn/6730096371?network=testnet. The fee payer address is `0x0a826f716ffe5bd551fdc31941d885d7cc9b050e1d12f13f2d9b81d3e8ba8440` but the explorer shows it as `0xa826...8440`

PR #1136, opened 2025-05-21
[account] Fix truncated addresses

gregnazario:
Ensures that we un-truncate truncated addresses in buttons.

Resolves https://github.com/aptos-labs/explorer/issues/1135


ying-w:
do you want to special case 0x0 - 0xa?

banool (approved):

banool:
It should already be special cased if we're using `AccountAddress.toString()`.

PR #1142, opened 2025-06-03
show fa migration banner

0xaptosj:

### DIFF
--- a/src/pages/LandingPage/CoinToFAMigrationBanner.tsx
+++ b/src/pages/LandingPage/CoinToFAMigrationBanner.tsx
@@ -29,7 +29,6 @@ export function CoinToFAMigrationBanner() {
     });
     window.open(
       `https://explorer.aptoslabs.com/account/${addressFromWallet(account?.address)}/coins`,
-      "_blank",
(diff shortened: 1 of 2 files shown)


src/pages/LandingPage/CoinToFAMigrationBanner.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const handleClick = () => {
+    setOpen(!open);
+    logEvent("coin_to_fa_migration_banner_clicked", null, {
+      wallet_address: addressFromWallet(account?.address),
+      wallet_name: wallet?.name ?? "",
+    });
+    window.open(
+      `https://explorer.aptoslabs.com/account/${addressFromWallet(account?.address)}/coins`,
+      "_blank",
```
gregnazario: This should not be _blank as it opens a new tab.

It should stay as the same tab so you don't connect you wallet again

0x-j: done

gregnazario (approved):

PR #1193, opened 2025-08-22
Add block time to analytics page

ying-w:
### Description

Block gap is a way to measure block production rate. For every day, this is the median time in ms between blocks

### DIFF
--- a/src/pages/Analytics/Charts/DailyBlockGapChart.tsx
+++ b/src/pages/Analytics/Charts/DailyBlockGapChart.tsx
@@ -28,7 +28,7 @@ export default function DailyBlockGapChart({
     <CardOutline>
       <ChartTitle
         label="Median Block Time"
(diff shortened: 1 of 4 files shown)


src/pages/Analytics/Charts/DailyBlockGapChart.tsx
```diff
@@ -0,0 +1,44 @@
...
+  days,
+}: DailyBlockGapChartProps) {
+  const labels = getLabels(data, days);
+  const dataset = getDataset(data, days);
+
+  return (
+    <CardOutline>
+      <ChartTitle
+        label="Block Time"
+        tooltip="Median block time (gap between blocks) in milliseconds. Will vary based on network latency between validators and blockchain congestion (more txn per block vs faster blocks)."
```
ibalajiarun: likely clarify that it's the median from all blocks in a day.

PR #1209, opened 2025-09-18
[rewards] Fix rewards rate from on-chain data

gregnazario:


ying-w (approved):
Thanks for updating staking rate to the correct value

PR #1214, opened 2025-09-25
Fix/coins fallback metadata

Zaran-12:
### Background
On local networks, v2 balances (`current_fungible_asset_balances`) may return `metadata = null`
for some assets (e.g. LBT: `0x1::libra2_coin::Libra2Coin`) while v1 `coin_infos` is not fully
ingested yet. The Coins tab filtered out rows with missing metadata, resulting in "No Data Found".

### Changes
- CoinsTab: do not drop rows with `metadata == null`; add **fallback metadata** builder
  - If asset_type is a TypeTag → treat as `v1`
  - Special-case LBT: name "Libra2 Coin", symbol "LBT", decimals 8
  - Compute USD value only when price is available; otherwise 0 / null
- (Optional) Hooks: restore/keep REST `CoinInfo<T>` fallback when indexer metadata is missing
- Small fixes around balance display/format (APT/LBT decimals)

### Why safe
Pure front-end rendering change. No impact on backend/indexer. When indexer metadata is present,
it takes precedence; fallback is only used for `metadata == null`.

### Testing
- Fullnode: `http://127.0.0.1:8080`
- Indexer GraphQL: `http://127.0.0.1:8090/v1/graphql`
- Ensure `CoinInfo<0x1::libra2_coin::Libra2Coin>` exists:
  curl -s "http://127.0.0.1:8080/v1/accounts/0x1/resource/0x1::coin::CoinInfo%3C0x1::libra2_coin::Libra2Coin%3E"
- Account page → Coins tab now shows LBT row even when v2 metadata is null.

### Screenshots
(please attach before/after if possible)

### Checklist
- [ ] Code builds locally
- [ ] `pnpm lint` / `yarn lint` / `npm run lint` pass locally
- [ ] No secrets or `.env.local` are included


gregnazario:
This does not look like it's a pull request for Aptos.  I'm closing it for now

PR #1226, opened 2025-10-08
[explorer] Shelbynet unhide

hariria:
Will wait to merge this until later.


src/constants.tsx
```diff
@@ -8,15 +9,15 @@ export const devnetUrl =
   "https://api.devnet.staging.aptoslabs.com/v1";
 
 export const networks: Record<string, string> = {
-  mainnet: "https://api.mainnet.aptoslabs.com/v1",
-  testnet: "https://api.testnet.staging.aptoslabs.com/v1",
+  mainnet: NetworkToNodeAPI[AptosNetwork.MAINNET],
+  testnet: NetworkToNodeAPI[AptosNetwork.TESTNET],
```
banool, line 13: This will cause issues because the key we use is for staging. Let's leave the URL as-is, it's good for dogfooding to use staging.

PR #1329, opened 2025-12-30
feat: improve view function result display with proper JSON formatting

WGB5445:
# Description

This PR improves the display of view function results in the Contract interaction panel. Previously, JSON results were displayed inline without proper formatting, making complex nested structures difficult to read and causing horizontal overflow issues.

# Changes

- Proper JSON formatting: Use `<pre>` element with `whiteSpace: pre-wrap` to preserve JSON indentation and line breaks
- Prevent overflow: Add `overflow: auto` and `maxHeight: 400px` to handle large JSON responses gracefully
- Monospace font: Use monospace font family for better code readability
Vertical layout: Changed from responsive row/column layout to consistent column layout for cleaner presentation
- Error styling: Errors now display in red color for better visibility
Code cleanup: Removed unused `isWideScreen` and `theme` variables

Before

image

Now

image

### DIFF
--- a/src/pages/Account/Tabs/ModulesTab/Contract.tsx
+++ b/src/pages/Account/Tabs/ModulesTab/Contract.tsx
@@ -695,6 +695,8 @@ function ReadContractForm({
   const aptosClient = useAptosClient();
   const sdkV2Client = useSdkV2Client();
   const [result, setResult] = useState<Types.MoveValue[]>();
+  const theme = useTheme();
+  const isWideScreen = useMediaQuery(theme.breakpoints.up("md"));
   const [errMsg, setErrMsg] = useState<string>();
   const [inProcess, setInProcess] = useState(false);
   const [formValid, setFormValid] = useState(false);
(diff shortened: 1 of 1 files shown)


src/pages/Account/Tabs/ModulesTab/Contract.tsx
```diff
@@ -695,8 +695,6 @@ function ReadContractForm({
   const aptosClient = useAptosClient();
   const sdkV2Client = useSdkV2Client();
   const [result, setResult] = useState<Types.MoveValue[]>();
-  const theme = useTheme();
-  const isWideScreen = useMediaQuery(theme.breakpoints.up("md"));
```
gregnazario, line 699: Interesting, why get rid of the existing theme stuff?

WGB5445: Originally, I thought it would be better to keep the same layout structure, but after taking another look, I decided to revert to the previous approach: 

- on desktop, the result and the copy button are on the same row
- on mobile, the copy button is placed below the result.

PR #1596, opened 2026-04-24
fix(api): attribute geomi AG keys via api-key header on Aptos API gateway

gregnazario:
## Summary

Investigation showed that the production explorer embeds `AG-*` (geomi) keys in the client bundle, but those keys were sent as `Authorization: Bearer <key>`. Against `https://api.mainnet.aptoslabs.com`, that form returns **401 Unauthorized: API key not found**, while the same key with the **`api-key`** header returns **200**. The gateway also accepts unauthenticated requests for many paths, so the app could still work while Geomi reported **zero compute units** because the key was never accepted.

## Changes

- Centralize gateway auth in `aptosGatewayApiKeyHeaders()` (`app/lib/aptosGatewayAuth.ts`) and use it from `createAptosClient` and `GlobalConfig` (v1 + v2 clients).
- Unit tests for the header helper; spec + changelog updates.
- **E2E** (`e2e/aptosGatewayAuth.spec.ts`): Playwright loads `/?network=mainnet`, records requests whose URL parses as `https://api.mainnet.aptoslabs.com` (exact hostname + scheme; avoids CodeQL incomplete-substring alerts), asserts at least one request includes `api-key` equal to `VITE_APTOS_MAINNET_API_KEY`, and asserts none use `Authorization: Bearer AG-*`. Skips when the env var is unset (local builds without keys). CI already sets this var to match the built bundle.

## Verification

- `pnpm fmt && pnpm lint && pnpm test --run`
- `pnpm exec playwright test e2e/aptosGatewayAuth.spec.ts` (with `VITE_APTOS_MAINNET_API_KEY` matching the preview build)

After deploy, Geomi usage for the mainnet frontend key should begin accruing again once traffic uses the new build.

[Slack Thread](https://aptos-org.slack.com/archives/C040LV2NWQ4/p1777072526710699?thread_ts=1777072526.710699&cid=C040LV2NWQ4)

<div><a href="https://cursor.com/agents/bc-0c98b6e1-05d8-5971-850d-330c504a9a98"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-0c98b6e1-05d8-5971-850d-330c504a9a98"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


banool (changes requested):
We don't look at the api-key header, the AI probably got confused because we allow anonymous requests. It added this header and thought it fixed it but it as actually ignored.

PR #1680, opened 2026-07-06
fix(validators): replay mixed delegation reward events

WGB5445:
## Summary
- replay validator `My Deposits` principals across both legacy `*Event` names and the newer suffixless delegation event names
- apply `WithdrawStake` when reconstructing pending-inactive principal and break same-version ties with `event_index`
- render computed zero rewards as `0 APT` and add a regression test covering an anonymized production history

## Root cause
The validator detail reward helper only recognized legacy delegation event names (`AddStakeEvent`, `UnlockStakeEvent`, `ReactivateStakeEvent`) and never replayed `WithdrawStake` back out of pending-inactive principal. Mixed-format histories therefore left stale principal behind and overstated the displayed reward earned amount.

## Proof
The regression test now uses an anonymized production history copied from the reported mixed-format sequence. It covers the exact legacy/current event mix, asserts the final unlock produces the expected pending-inactive principal, and then asserts the final withdraw returns that principal to zero.

## Validation
- `CI=true pnpm test --run app/pages/DelegatoryValidator/utils.test.ts`
- `CI=true pnpm lint`
- `CI=true pnpm test --run`

## Linear
- [DVR-178](https://linear.app/aptoslabs/issue/DVR-178/fix-validator-my-deposits-reward-replay-for-mixed-delegation-event)


app/pages/DelegatoryValidator/utils.tsx
```diff
@@ -73,29 +102,30 @@ export function getStakeOperationPrincipals(activities: {
...
           pendingInactivePrincipals += amount;
           break;
-        case "ReactivateStakeEvent":
+        case "ReactivateStake":
           activePrincipals += amount;
           pendingInactivePrincipals -= amount;
           break;
+        case "WithdrawStake":
+          pendingInactivePrincipals -= amount;
+          break;
```
gregnazario, line 131: ah it's handling both types now

app/pages/DelegatoryValidator/utils.tsx
```diff
@@ -29,6 +29,34 @@ export type StakePrincipals = {
...
+export function getStakeRewardsEarned(
+  stake: Types.MoveValue,
+  principalsAmount: number | undefined,
+) {
+  if (principalsAmount === undefined) {
+    return undefined;
+  }
+
+  return Math.max(Number(stake) - principalsAmount, 0);
+}
```
gregnazario, line 42: Stake from the indexer will be an int64 (with no negatives), but number is only 2^53 -1 instead of 2^64 -1

So, you should use `bigint` for the math.

gregnazario (approved):

PR #1757, opened 2026-09-17
fix(i18n): rename inlineMarkup parser to avoid macOS case collision

WGB5445:
## Summary

On case-insensitive APFS (macOS), `InlineMarkup.tsx` and `inlineMarkup.ts` resolve to the same module. Imports of `./InlineMarkup` therefore got the parser instead of the React component (`undefined` export), SSR threw "Element type is invalid", and `pnpm dev` returned 500.

Rename the parser to `parseInlineMarkup.ts` (and the matching test file) so the paths no longer collide.

## Test plan

- [x] `pnpm test --run app/i18n/parseInlineMarkup.test.ts`
- [ ] Manual: `pnpm dev` on macOS — home/`/guide` SSR loads without 500

## Scope

This PR is **only** the macOS case-collision rename. Guide layout and network-tooltip work are intentionally not included.


gregnazario (approved):
you should always make your macOS case sensitive

case insensitivity makes so many problems

WGB5445:
> you should always make your macOS case sensitive
> 
> case insensitivity makes so many problems

Yeah, I agree, ideally we'd go with case sensitivity so it matches the Linux default. 
But Mac and Windows are case-insensitive by default. 

I'd still prefer it if every machine could just work for development without needing too much custom setup.

